PR #24, opened 2023-10-23
Log version info on application start

korotkov-aerospike:


korotkov-aerospike:
Just to be sure we are actually running the desired version

reugn (approved):

PR #25, opened 2023-10-23
Improve OpenApi endpoint annotations

korotkov-aerospike:
Add `@ID` annotation and response codes


reugn (commented):
Format the annotations to match the original indentation.

korotkov-aerospike:
ha, used tabs, so in my IDE it looked nice

PR #32, opened 2023-10-26
APPS-935 Reduce the run time of docker build

edramaspk:


buildx-image-creation.sh
```diff
@@ -0,0 +1,11 @@
+TAG_LATEST=${TAG_LATEST:-false}
```
davi17g, line 1:
```suggestion
#!/bin/bash
TAG_LATEST=${TAG_LATEST:-false}
```

dockerfiles/create-abs-image.sh
```diff
@@ -0,0 +1,12 @@
+TAG_LATEST=${TAG_LATEST:-false}
```
davi17g, line 1:
```suggestion
#!/bin/bash
TAG_LATEST=${TAG_LATEST:-false}
```

dockerfiles/create-base-image.sh
```diff
@@ -0,0 +1,5 @@
+docker buildx create --use --name mybuilder
```
davi17g, line 1:
```suggestion
#!/bin/bash
docker buildx create --use --name mybuilder
```

edramaspk: Done.

buildx-image-creation.sh
reugn, line 1: Why is this script required?

reugn (approved):

PR #36, opened 2023-10-31
Add password path property to the AerospikeCluster configuration

reugn:


pkg/model/config.go
```diff
@@ -54,11 +57,39 @@ type RateLimiterConfig struct {
 
 // AerospikeCluster represents the configuration for an Aerospike cluster for backup.
 type AerospikeCluster struct {
-	Name     *string `yaml:"name,omitempty" json:"name,omitempty"`
-	Host     *string `yaml:"host,omitempty" json:"host,omitempty"`
-	Port     *int    `yaml:"port,omitempty" json:"port,omitempty"`
-	User     *string `yaml:"user,omitempty" json:"user,omitempty"`
-	Password *string `yaml:"password,omitempty" json:"password,omitempty"`
+	pwdOnce      sync.Once
```
korotkov-aerospike, line 60: why without serialization annotations? We need to be able to provide password

korotkov-aerospike (approved):

PR #43, opened 2023-11-02
APPS-974 Fix backup size computation

reugn:


korotkov-aerospike (approved):

korotkov-aerospike:
We can use the same logic to get file size for s3 objects, what do you think?
I think, for users it would be more convenient to have it everywhere

PR #44, opened 2023-11-05
APPS-963 Use array to store request's bin list

korotkov-aerospike:


pkg/model/config.go
```diff
@@ -121,8 +121,8 @@ type BackupPolicy struct {
 	Namespace          *string     `yaml:"namespace,omitempty" json:"namespace,omitempty"`
 	Parallel           *int32      `yaml:"parallel,omitempty" json:"parallel,omitempty"`
 	SetList            *[]string   `yaml:"set_list,omitempty" json:"set_list,omitempty"`
+	BinList            *[]string   `yaml:"bin_list,omitempty" json:"bin_list,omitempty"`
 	NodeList           *string     `yaml:"node_list,omitempty" json:"node_list,omitempty"`
```
reugn, line 125: Make it an array as well, use the `[]string` type for all of the array fields

pkg/model/config.go
reugn, line 1: * `PartitionList` can also be an array. Check whether partition ranges are supported.

korotkov-aerospike: No,  unfortunately

`List of partitions to back up. Partition filters can be ranges, individual partitions, or records after a specific digest within a single partition`

pkg/shared/backup.go
```diff
@@ -22,6 +22,8 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/aws/smithy-go/ptr"
```
reugn, line 25: Move this import to the bottom

pkg/model/config.go
```diff
@@ -120,9 +120,9 @@ type BackupPolicy struct {
 	Storage            *string     `yaml:"storage,omitempty" json:"storage,omitempty"`
 	Namespace          *string     `yaml:"namespace,omitempty" json:"namespace,omitempty"`
 	Parallel           *int32      `yaml:"parallel,omitempty" json:"parallel,omitempty"`
-	SetList            *[]string   `yaml:"set_list,omitempty" json:"set_list,omitempty"`
-	NodeList           *string     `yaml:"node_list,omitempty" json:"node_list,omitempty"`
-	BinList            *string     `yaml:"bin_list,omitempty" json:"bin_list,omitempty"`
+	SetList            []string    `yaml:"set_list,omitempty" json:"set_list,omitempty"`
+	BinList            []string    `yaml:"bin_list,omitempty" json:"bin_list,omitempty"`
+	NodeList           []string    `yaml:"node_list,omitempty" json:"node_list,omitempty"`
```
reugn, line 125: Use this field in the backup configuration.

korotkov-aerospike: I think, it would be better to have structure for Node

reugn (approved):

PR #57, opened 2023-11-21
APPS-1003 Filter backups in progress

korotkov-aerospike:


pkg/service/backup_backend_s3.go
```diff
@@ -42,33 +45,39 @@ func (s *BackupBackendS3) writeState(state *model.BackupState) error {
...
-			return []model.BackupDetails{{
-				Key:          ptr.String(s3prefix + backupFolder),
-				LastModified: &s.readState().LastRun,
-				Size:         ptr.Int64(s.dirSize(backupFolder)),
-			}}, nil
+		if len(files) == 0 {
+			return []model.BackupDetails{}, nil
 		}
-		return []model.BackupDetails{}, nil
+		if s.fullBackupInProgress.Load() {
```
reugn, line 55: Can be combined with the previous if.

pkg/service/backup_backend_s3.go
```diff
@@ -42,33 +45,39 @@ func (s *BackupBackendS3) writeState(state *model.BackupState) error {
 func (s *BackupBackendS3) FullBackupList() ([]model.BackupDetails, error) {
 	backupFolder := s.Path + "/" + model.FullBackupDirectory + "/"
 	s3prefix := "s3://" + s.bucket
+	lastRun := s.readState().LastRun
```
reugn, line 48: Init this value close to the use place

pkg/service/backup_backend_s3.go
```diff
@@ -11,20 +12,22 @@ import (
 // saving state to AWS S3.
 type BackupBackendS3 struct {
 	*S3Context
-	stateFilePath string
-	backupPolicy  *model.BackupPolicy
+	stateFilePath        string
+	backupPolicy         *model.BackupPolicy
+	fullBackupInProgress *atomic.Bool
```
reugn, line 17: Make it non-pointer and no need to pass it in the constructor.

pkg/service/backup_backend_s3.go
```diff
@@ -86,19 +95,22 @@ func (s *BackupBackendS3) dirSize(path string) int64 {
 
 // IncrementalBackupList returns a list of available incremental backups.
 func (s *BackupBackendS3) IncrementalBackupList() ([]model.BackupDetails, error) {
+	lastIncrRun := s.readState().LastIncrRun
```
reugn, line 98: Init it right before the for-range loop

pkg/service/backup_backend_local.go
```diff
@@ -17,22 +18,25 @@ import (
 // BackupBackendLocal implements the BackupBackend interface by
 // saving state to the local file system.
 type BackupBackendLocal struct {
-	path          string
-	stateFilePath string
-	backupPolicy  *model.BackupPolicy
+	path                 string
+	stateFilePath        string
+	backupPolicy         *model.BackupPolicy
+	fullBackupInProgress *atomic.Bool
```
reugn, line 24: Make it non-pointer and no need to pass it in the constructor.

pkg/service/backup_backend_local.go
```diff
@@ -78,21 +82,28 @@ func (local *BackupBackendLocal) FullBackupList() ([]model.BackupDetails, error)
...
-		if len(entries) > 0 {
-			return []model.BackupDetails{{
-				Key:          ptr.String(backupFolder),
-				LastModified: &local.readState().LastRun,
-			}}, nil
+		if len(entries) == 0 {
+			return []model.BackupDetails{}, nil
 		}
-		return []model.BackupDetails{}, nil
+		if local.fullBackupInProgress.Load() {
```
reugn, line 91: Combine it with the previous if.

pkg/service/backup_backend_s3.go
korotkov-aerospike: it's different scenarios, so I prefer to keep them separate

pkg/service/backup_backend_s3.go
korotkov-aerospike: That's the closest.
it is used both inside of the if and after it

pkg/service/backup_backend_s3.go
korotkov-aerospike: That's exactly why I use pointer: I need same value in BackupBackend and BackupHandler.

reugn (approved):

PR #58, opened 2023-11-22
Use backup routine name in logs

korotkov-aerospike:


korotkov-aerospike:
We should print routine name, not policy in the logs.
Also, I want the version to be printed in any case.

PR #73, opened 2023-12-10
APPS-1034 Delete incremental backups

korotkov-aerospike:


pkg/service/s3_context.go
```diff
@@ -159,26 +159,33 @@ func removeLeadingSlash(s string) string {
 }
 
 // CleanDir cleans the directory with the given name.
-func (s *S3Context) CleanDir(name string) {
-	path := s.Path + "/" + name
+func (s *S3Context) CleanDir(name string) error {
+	path := removeLeadingSlash(s.Path + "/" + name)
```
korotkov-aerospike, line 163: This line is actually fix

reugn: Needs to be tested on both MinIO and S3.

korotkov-aerospike: In the ecosystem project

korotkov-aerospike:
I tested on AWS.

pkg/service/s3_context_test.go
```diff
@@ -0,0 +1,32 @@
+//go:build !ci
+
+package service
+
+import (
+	"github.com/aerospike/backup/pkg/model"
+	"github.com/aws/smithy-go/ptr"
+	"testing"
```
reugn, line 8: Format imports.

reugn (approved):

PR #74, opened 2023-12-11
APPS-1036 Always calculate backup size property

korotkov-aerospike:


pkg/service/backup_backend_local.go
```diff
@@ -163,29 +164,32 @@ func toBackupDetails(e fs.DirEntry, prefix string) model.BackupDetails {
 	dirInfo, err := e.Info()
 	if err == nil {
 		details.LastModified = util.Ptr(dirInfo.ModTime())
-		details.Size = util.Ptr(dirEntrySize(prefix, e, dirInfo))
+		details.Size = folderSize(prefix)
```
reugn, line 167: You can take it out of the if statement, since it is not related to the `dirInfo` anymore.

pkg/service/backup_backend_local.go
```diff
@@ -163,29 +164,32 @@ func toBackupDetails(e fs.DirEntry, prefix string) model.BackupDetails {
...
+				return err
 			}
+			size += info.Size()
 		}
-		return totalSize
+		return nil
+	})
+
+	if err != nil {
+		slog.Error("failed to calculate size", "path", path)
```
reugn, line 190: Add `"err", err` to the log

reugn (approved):

PR #76, opened 2023-12-12
APPS-1030 Add read write mutex to state file

korotkov-aerospike:


pkg/service/backup_backend_local.go
```diff
@@ -164,10 +170,11 @@ func toBackupDetails(e fs.DirEntry, prefix string) model.BackupDetails {
 	if err == nil {
 		lastModified = util.Ptr(dirInfo.ModTime())
 	}
+	path := filepath.Join(prefix, e.Name())
 	return model.BackupDetails{
-		Key:          util.Ptr(filepath.Join(prefix, e.Name())),
+		Key:          util.Ptr(path),
 		LastModified: lastModified,
-		Size:         folderSize(prefix),
```
korotkov-aerospike, line 170: That was the error with file size: we calculated correctly, but for the wrong folder

pkg/service/backup_backend_local.go
```diff
@@ -23,6 +24,7 @@ type BackupBackendLocal struct {
 	stateFilePath        string
 	backupPolicy         *model.BackupPolicy
 	fullBackupInProgress *atomic.Bool // BackupBackend needs to know if full backup is running to filter it out
+	statFileMutex        sync.RWMutex
```
reugn, line 27: stateFileMutex?

pkg/service/backup_backend_local.go
```diff
@@ -55,6 +57,8 @@ func prepareDirectory(path string) {
 }
 
 func (local *BackupBackendLocal) readState() *model.BackupState {
+	local.statFileMutex.RLock()
+	defer local.statFileMutex.RUnlock()
```
reugn, line 61: You can unlock right after the `os.ReadFile` operation.

pkg/service/backup_backend_s3.go
```diff
@@ -15,6 +16,7 @@ type BackupBackendS3 struct {
 	stateFilePath        string
 	backupPolicy         *model.BackupPolicy
 	fullBackupInProgress *atomic.Bool // BackupBackend needs to know if full backup is running to filter it out
+	statFileMutex        sync.RWMutex
```
reugn, line 19: See the previous naming comment.

reugn (approved):

PR #81, opened 2023-12-13
APPS-1042 Add support for Secret Agent configuration

reugn:


pkg/model/backup_routine.go
```diff
@@ -7,9 +7,10 @@ import (
 
 // BackupRoutine represents a scheduled backup operation routine.
 type BackupRoutine struct {
-	BackupPolicy  string `yaml:"backup-policy,omitempty" json:"backup-policy,omitempty"`
-	SourceCluster string `yaml:"source-cluster,omitempty" json:"source-cluster,omitempty"`
-	Storage       string `yaml:"storage,omitempty" json:"storage,omitempty"`
+	BackupPolicy  string  `yaml:"backup-policy,omitempty" json:"backup-policy,omitempty"`
+	SourceCluster string  `yaml:"source-cluster,omitempty" json:"source-cluster,omitempty"`
+	Storage       string  `yaml:"storage,omitempty" json:"storage,omitempty"`
+	SecretAgent   *string `yaml:"secret-agent,omitempty" json:"secret-agent,omitempty"`
```
korotkov-aerospike, line 13: lets have it just a string, like other fields?

pkg/service/backup_scheduler.go
```diff
@@ -229,7 +235,8 @@ func (h *BackupHandler) runIncrementalBackup(now time.Time) {
 		opts.ModAfter = &lastIncrRunEpoch
 		started := time.Now()
 		slog.Debug("Starting incremental backup", "name", h.routineName)
-		stats = backupService.BackupRun(h.backupRoutine, h.backupIncrPolicy, h.cluster, h.storage, opts)
+		stats = backupService.BackupRun(h.backupRoutine, h.backupIncrPolicy, h.cluster,
```
korotkov-aerospike, line 238: we have so many params here, maybe just pass the whole 'h', WDYT?

korotkov-aerospike (approved):

pkg/service/backup_scheduler.go
reugn: Right, I've seen that amount of parameters growing. But passing in a BackupHandler doesn't feel the right thing to do. Let's leave this for another refactoring cycle.

PR #86, opened 2023-12-19
APPS-1054 Fix files deletion on S3

korotkov-aerospike:


pkg/service/s3_context.go
```diff
@@ -33,11 +33,7 @@ func NewS3Context(storage *model.Storage) *S3Context {
 	// Load the SDK's configuration from environment and shared config, and
 	// create the client with this.
 	ctx := context.TODO()
-	cfg, err := config.LoadDefaultConfig(
-		ctx,
-		config.WithSharedConfigProfile(*storage.S3Profile),
-		config.WithRegion(*storage.S3Region),
```
korotkov-aerospike, line 39: profile is an optional field, we don't need it for aws

pkg/service/s3_context.go
```diff
@@ -76,6 +72,17 @@ func NewS3Context(storage *model.Storage) *S3Context {
...
 }
 
+func createConfig(ctx context.Context, storage *model.Storage) (aws.Config, error) {
+	cfgOptions := []func(*config.LoadOptions) error{
+		config.WithRegion(*storage.S3Region),
+	}
+	if storage.S3Profile != nil {
+		cfgOptions = append(cfgOptions, config.WithSharedConfigProfile(*storage.S3Profile))
+	}
+	cfg, err := config.LoadDefaultConfig(ctx, cfgOptions...)
```
reugn, line 82: return here the result of the function call

reugn (approved):

PR #88, opened 2023-12-20
APPS-1047 Add time filters to the list of full backups endpoint

reugn:


pkg/service/backup_backend_local.go
```diff
@@ -111,7 +111,9 @@ func (local *BackupBackendLocal) FullBackupList() ([]model.BackupDetails, error)
 	for _, e := range entries {
 		if e.IsDir() {
 			details := toBackupDetails(e, backupFolder)
-			if details.LastModified.Before(lastRun) {
+			if details.LastModified.Before(lastRun) &&
```
korotkov-aerospike, line 114: In the  RemoveFiles branch, when we have only one backup maximum, probably should filter it too, to be consistent.

pkg/service/backup_backend_s3.go
```diff
@@ -88,6 +95,21 @@ func (s *BackupBackendS3) FullBackupList() ([]model.BackupDetails, error) {
 	return result, err
 }
 
+// filterFolders returns backup folders that were created in the specified time range.
+func filterFolders(folders []types.CommonPrefix, from, to int64) ([]types.CommonPrefix, error) {
```
korotkov-aerospike, line 99: We have method s.GetTime, that fetches time from S3 (and cache it)
Here you fetch time based on folder name.
I think, we should use the same approach in both places.

reugn: I don't want to make this extra API call when pre-filtering.

internal/server/backup_handlers.go
```diff
@@ -13,12 +15,22 @@ import (
 // @Tags     Backup
 // @Produce  json
 // @Param    name query string false "Backup routine name"
+// @Param    from query int false "Lower bound timestamp filter"
```
korotkov-aerospike, line 18: Should've mark it as int64, but I could not find way to do it

reugn: There is no int64/long type in Swagger. int64 will result in the integer type, too. Do you encounter type incompatibility issues generating the client?

korotkov-aerospike: I found it!

pkg/service/backup_backend_local.go
```diff
@@ -100,6 +100,10 @@ func (local *BackupBackendLocal) FullBackupList() ([]model.BackupDetails, error)
 		if local.fullBackupInProgress.Load() {
 			return []model.BackupDetails{}, nil
 		}
+		// check request time boundaries
+		if lastRun.UnixMilli() < from || lastRun.UnixMilli() >= to {
```
korotkov-aerospike, line 104: same for backupbackendS3

korotkov-aerospike (approved):

PR #99, opened 2023-12-26
APPS-1067 Update golang dependencies

reugn:


pkg/service/backup_backend_s3.go
```diff
@@ -139,7 +139,7 @@ func (s *BackupBackendS3) IncrementalBackupList() ([]model.BackupDetails, error)
 		details := model.BackupDetails{
 			Key:          ptr.String(s3prefix + "/" + *object.Key),
 			LastModified: object.LastModified,
-			Size:         &object.Size,
+			Size:         object.Size,
```
korotkov-aerospike, line 142: You need to update size field itself too

korotkov-aerospike (approved):

PR #114, opened 2024-01-17
APPS-1081 Add ability to do a full backup ad-hoc

reugn:


internal/server/backup_handlers.go
```diff
@@ -109,3 +112,44 @@ func parseTimestamp(value string, defaultValue int64) (int64, error) {
...
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	}
+	if delayMillis < 0 {
+		http.Error(w, "nonpositive delay query parameter", http.StatusBadRequest)
+		return
+	}
+	job := service.GetFullBackupJobByRoutineName(routineName)
```
korotkov-aerospike, line 140: I think, we should move all the actual business logic to scheduler (or even new file in it's package) and leave in handler only parameter validation and execution of those extracted code

korotkov-aerospike (approved):

PR #115, opened 2024-01-17
APPS-1100 Filter out backups without metadata

korotkov-aerospike:


pkg/service/s3_context.go
```diff
@@ -192,12 +192,12 @@ func (s *S3Context) CleanDir(name string) error {
 	return s.DeleteFolder(path)
 }
 
-func (s *S3Context) GetMetadata(l types.CommonPrefix) *model.BackupMetadata {
+func (s *S3Context) GetMetadata(l types.CommonPrefix) (*model.BackupMetadata, error) {
 	metadata, err := s.metadataCache.Get(*l.Prefix)
-	if err == nil {
-		return metadata.(*model.BackupMetadata)
+	if err != nil {
+		return &model.BackupMetadata{}, err
```
reugn, line 198: You can return here
```go
return nil, err
```

reugn (approved):

PR #116, opened 2024-01-18
APPS-1101 Do not warn if state file doesn't exist

reugn:


korotkov-aerospike:
Need to do same for local backups
pkg/service/backup_backend_local.go:59

korotkov-aerospike (approved):

PR #119, opened 2024-01-21
APPS-1070 Make use of S3LogLevel field in storage

reugn:


pkg/model/storage.go
```diff
@@ -32,6 +35,8 @@ const (
 	S3
 )
 
+var validS3LogLevels = []string{"OFF", "FATAL", "ERROR", "WARN", "INFO", "DEBUG", "TRACE"}
```
korotkov-aerospike, line 38: We need to tell endusers what are the valid levels. Maybe add them to error message and/or swagger doc?

PR #126, opened 2024-02-01
APPS-1114 Add option to keep incremental backups

korotkov-aerospike:


pkg/service/backup_handler.go
```diff
@@ -223,15 +228,15 @@ func (h *BackupHandler) writeState() {
 
 func getFullPath(storage *model.Storage, backupPolicy *model.BackupPolicy, namespace string, now time.Time) *string {
 	if backupPolicy.RemoveFiles != nil && !*backupPolicy.RemoveFiles {
-		path := fmt.Sprintf("%s/%s/%s/%s", *storage.Path, model.FullBackupDirectory, timeSuffix(now), namespace)
+		path := fmt.Sprintf("%s/%s/%s/%s/", *storage.Path, model.FullBackupDirectory, timeSuffix(now), namespace)
```
reugn, line 231: Was this tested with both S3 and MinIO?

internal/server/backup_handlers.go
```diff
@@ -92,6 +86,24 @@ func (ws *HTTPServer) getAvailableBackups(
 	}
 }
 
+func parseTimebounds(r *http.Request) (int64, int64, error) {
```
reugn, line 89: nit: It would make sense to return a pair struct containing the time bounds.

korotkov-aerospike:
Tests passed
http://3.65.168.247:8080/ui/#pitr/launches/all/5153

reugn (approved):

PR #127, opened 2024-02-02
APPS-1113 Refactor cluster configuration structure

reugn:


arrowplum:
Does this validate against the schema? I think no, right? It needs "aerospike-clusters" instead of "aerospike"

korotkov-aerospike (approved):

PR #129, opened 2024-02-06
APPS-1125 Prefer rack flag

korotkov-aerospike:


pkg/model/backup_routine.go
```diff
@@ -29,6 +29,8 @@ type BackupRoutine struct {
 	BinList []string `yaml:"bin-list,omitempty" json:"bin-list,omitempty"`
 	// The list of nodes in the Aerospike cluster to run the backup for.
 	NodeList []Node `yaml:"node-list,omitempty" json:"node-list,omitempty"`
+	// A list of Aerospike Server rack IDs to prefer when reading records for a backup.
+	PreferRacks []uint `yaml:"prefer-racks,omitempty" json:"prefer-racks,omitempty"`
```
reugn, line 33: It's better to use int32 here and validate for negative value.

reugn (approved):

PR #132, opened 2024-02-07
APPS-1127 Update config endpoints

korotkov-aerospike:


pkg/model/storage.go
```diff
@@ -13,7 +13,7 @@ import (
 // @Description Storage represents the configuration for a backup storage details.
 type Storage struct {
 	// The type of the storage provider (0 - Local, 1 - AWS S3).
-	Type StorageType `yaml:"type,omitempty" json:"type,omitempty"`
+	Type StorageType `yaml:"type" json:"type"`
```
korotkov-aerospike, line 16: I want type = 0 to be serialized, not skipped.

internal/server/server.go
```diff
@@ -280,3 +284,9 @@ func (ws *HTTPServer) configRoutineActionHandler(w http.ResponseWriter, r *http.
 		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
 	}
 }
+
+// getLastURLSegment extracts the last segment (path parameter) from the url path
+func getLastURLSegment(urlPath string) string {
+	urlParts := strings.Split(urlPath, "/")
+	return urlParts[len(urlParts)-1]
+}
```
reugn, line 292:
```suggestion
// getLastUrlSegment extracts the name path parameter for the configuration get endpoints.
// TODO: replace with the PathValue accessor after upgrading to go1.22.
func getLastUrlSegment(urlPath string) string {
	urlParts := strings.SplitAfterN(urlPath, "/", 4)
	return urlParts[len(urlParts)-1]
}
```

internal/server/server.go
```diff
@@ -280,3 +284,10 @@ func (ws *HTTPServer) configRoutineActionHandler(w http.ResponseWriter, r *http.
 		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
 	}
 }
+
+// getLastUrlSegment extracts the name path parameter for the configuration get endpoints.
+// TODO: replace with the PathValue accessor after upgrading to go1.22.
+func getLastUrlSegment(urlPath string) string {
```
reugn, line 290:
```suggestion
// getLastURLSegment extracts the name path parameter for the configuration get endpoints.
// TODO: replace with the PathValue accessor after upgrading to go1.22.
func getLastURLSegment(urlPath string) string {
```

reugn (approved):

PR #148, opened 2024-02-19
APPS-1144 Add encryption support

reugn:


pkg/model/encryption_policy.go
```diff
@@ -0,0 +1,40 @@
...
+	EncryptNone   = "NONE"
+	EncryptAES128 = "AES128"
+	EncryptAES256 = "AES256"
+)
+
+// EncryptionPolicy contains backup encryption information.
+// @Description EncryptionPolicy contains backup encryption information.
+type EncryptionPolicy struct {
+	// The encryption mode to be used (NONE, AES128, AES256)
+	Mode string `yaml:"mode,omitempty" json:"mode,omitempty" default:"NONE"`
```
korotkov-aerospike, line 19:
```suggestion
	Mode string `yaml:"mode,omitempty" json:"mode,omitempty" default:"NONE" enums:"NONE,AES128,AES256"`
```

That would add enums to openapi (and java client)

pkg/shared/util.go
```diff
@@ -87,3 +103,35 @@ func setTLSOptions(tlsName **C.char, tlsConfig *C.as_config_tls, tls *model.TLS)
...
+		switch policy.Mode {
+		case model.EncryptNone:
+			*encryptMode = C.IO_PROXY_ENCRYPT_NONE
+		case model.EncryptAES128:
+			*encryptMode = C.IO_PROXY_ENCRYPT_AES128
+		case model.EncryptAES256:
+			*encryptMode = C.IO_PROXY_ENCRYPT_AES256
+		}
+		if policy.KeyFile != nil {
+			*pKey = C.enc_key_malloc()
```
korotkov-aerospike, line 120: do we need to free this memory?

reugn: It is checked for null and freed in the config destroy functions.

korotkov-aerospike (approved):

PR #151, opened 2024-02-20
APPS-1151 Add validation for routine on creation

korotkov-aerospike:


korotkov-aerospike:
I've moved validation of routine's integrity into it's Validate method and call it both on app load and routine's update.
Also, unified validation errors.

reugn (approved):

PR #155, opened 2024-02-25
APPS-1159 Use unsigned Integer to store backup and restore statistics

korotkov-aerospike:


korotkov-aerospike:
I've updated automation too (to use Long in Java, not Integer), test run
http://3.65.168.247:8080/ui/#pitr/launches/all/5357

reugn (approved):

PR #160, opened 2024-02-28
Fix usage of sealed flag for full backup

korotkov-aerospike:


korotkov-aerospike:
I found an issue during testing.

reugn (approved):

PR #166, opened 2024-03-10
Add a nightly build workflow for packages

davi17g:


.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Check if the last commit was today
+        id: check-commit
+        run: |
+          LAST_COMMIT_DATE="$(git log -1 --format=%cd --date=short)"
+          TODAY="$(date -u +%Y-%m-%d)"
+          if [ "$LAST_COMMIT_DATE" == "$TODAY" ]; then
+            echo "commit_today=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "commit_today=false" >> "$GITHUB_OUTPUT"
+          fi
```
reugn, line 39: You can extract those steps into a separate job and have a single if condition for the `nightly-build` job.
You will need to check out the code in both of them.

.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,101 @@
+name: Nightly Build
+
+on:
+
+  schedule:
+    - cron: '0 20 * * *'
```
reugn, line 6:
```suggestion
on:
  workflow_dispatch:
  schedule:
    - cron: '0 20 * * *'

```

reugn: Leave the option to trigger manually.

reugn (approved):

PR #167, opened 2024-03-13
APPS-1167 Backup configuration

korotkov-aerospike:


pkg/service/aerospoike_service.go
```diff
@@ -2,6 +2,11 @@ package service
 
 import (
 	"fmt"
+	"github.com/aerospike/aerospike-management-lib/asconfig"
```
reugn, line 5: Format imports.

pkg/service/restore_memory.go
```diff
@@ -198,6 +199,29 @@ func (r *RestoreMemory) findIncrementalBackupsForNamespace(
...
+	if !found {
+		return nil, fmt.Errorf("backend '%s' not found for restore", routine)
+	}
+	fullBackups, err := r.findLastFullBackup(backend, toTimeMillis)
+	if err != nil {
+		return nil, fmt.Errorf("last full backup not found: %v", err)
+	}
+
+	// fullBackups has backups for multiple namespaces, but same timestamp, they share same configuration.
+	lastFullBackup := fullBackups[0]
```
reugn, line 213: Check `fullBackups` is not empty.

pkg/service/backup_handler.go
```diff
@@ -96,9 +96,28 @@ func (h *BackupHandler) runFullBackupInternal(now time.Time) error {
...
+	infos := getClusterConfiguration(h.cluster)
+	if len(infos) == 0 {
+		slog.Warn("Could not read aerospike configuration", "name", h.routineName)
+		return
+	}
+	path := getConfigurationPath(h.backend.fullBackupsPath, now)
+	h.backend.CreateFolder(path)
+	for i, info := range infos {
+		confFilePath := fmt.Sprintf("%s/aerospike_%d.conf", path, i)
+		slog.Info("Write aerospike configuration", "path", confFilePath)
```
reugn, line 113: Better to be `Debug`?

internal/server/server.go
```diff
@@ -195,6 +195,8 @@ func (ws *HTTPServer) Start() {
 	// Restore job status endpoint
 	mux.HandleFunc(ws.api("/restore/status/{jobId}"), ws.restoreStatusHandler)
 
+	mux.HandleFunc(ws.api("/restore/configuration/{name}/{timestamp}"), ws.restoreConfig)
```
reugn, line 198: Add a comment explaining the endpoint. Also, having it under restore is misleading, since it returns data to the user.

pkg/service/aerospoike_service.go
```diff
@@ -2,8 +2,14 @@ package service
...
 import (
 	"fmt"
+	"log/slog"
 	"strings"
 
+	"github.com/aerospike/aerospike-management-lib/asconfig"
+	"github.com/aerospike/aerospike-management-lib/info"
+	"github.com/aerospike/backup/pkg/util"
+	"github.com/go-logr/logr"
+
```
reugn, line 12: Remove that new line and format.

internal/server/server.go
korotkov-aerospike: Yeah, but is it worth creating new category just for a single endpoint? We have everything about accessing backups under /restore

pkg/service/restore_memory.go
korotkov-aerospike: findLastFullBackup will return error if nothing is found, but I'll add check anyway.

reugn (approved):

PR #170, opened 2024-03-21
APPS-1173 Apply updated config without service restart

korotkov-aerospike:


pkg/service/backup_backends_holder.go
```diff
@@ -0,0 +1,60 @@
...
+	b.RLock()
+	defer b.RUnlock()
+	backend, found := b.data[name]
+	return backend, found
+}
+
+func (b *BackendHolderImpl) Get(name string) (*BackupBackend, bool) {
+	b.RLock()
+	defer b.RUnlock()
+	backend, found := b.data[name]
```
reugn, line 43: You just can `return b.data[name]`

pkg/service/backup_backends_holder.go
```diff
@@ -0,0 +1,60 @@
...
+	return backend, found
+}
+
+func NewBackupBackends(config *model.Config) *BackendHolderImpl {
+	return &BackendHolderImpl{
+		RWMutex: sync.RWMutex{},
```
reugn, line 49: This default assignment can be dropped.

pkg/service/backup_backends_holder.go
```diff
@@ -0,0 +1,60 @@
...
+		data:    BuildBackupBackends(config),
+	}
+}
+
+func BuildBackupBackends(config *model.Config) map[string]*BackupBackend {
+	backends := map[string]*BackupBackend{}
```
reugn, line 55: It's better to pre-allocate here.

internal/server/config_handlers.go
```diff
@@ -59,3 +59,20 @@ func (ws *HTTPServer) updateConfig(w http.ResponseWriter, r *http.Request) {
...
+// @ID          applyConfig
+// @Tags        Configuration
+// @Router      /v1/config/apply [post]
+// @Accept      json
+// @Success     200
+// @Failure     400 {string} string
+func (ws *HTTPServer) applyConfig(w http.ResponseWriter, _ *http.Request) {
+	err := service.ApplyNewConfig(ws.scheduler, ws.config, ws.backupBackends)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
reugn, line 74: Not sure `http.StatusBadRequest` is the appropriate code here.

pkg/service/backup_backends_holder.go
```diff
@@ -0,0 +1,60 @@
...
+	defer b.Unlock()
+	b.data = backends
+}
+
+var _ BackendsHolder = (*BackendHolderImpl)(nil)
+
+func (b *BackendHolderImpl) GetReader(name string) (BackupListReader, bool) {
+	b.RLock()
+	defer b.RUnlock()
+	backend, found := b.data[name]
```
reugn, line 36: Same as in the Get method.

reugn (approved):

PR #172, opened 2024-03-26
APPS-1176 Do not store default values when updating configuration file

reugn:


reugn:
Models must have exported fields for JSON encoding/decoding. It doesn't make sense to add getters to every public property without a reason.

korotkov-aerospike (approved):

PR #173, opened 2024-03-26
APPS-1177 Retry read password

korotkov-aerospike:


pkg/model/aerospike_cluster_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"testing"
+
+	"github.com/aws/smithy-go/ptr"
+	"github.com/stretchr/testify/assert"
+)
+
+const testdataFolder = "./testdata"
+const passwordPath = "./testdata/password.txt"
+
+func CreateValidFile() {
```
reugn, line 14: Can be private.

pkg/model/aerospike_cluster_test.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/aws/smithy-go/ptr"
+	"github.com/stretchr/testify/assert"
+)
+
+const testdataFolder = "./testdata"
+const passwordPath = "./testdata/password.txt"
```
reugn, line 12: Group those two.

reugn (approved):

PR #177, opened 2024-03-31
APPS-1184 Fix remote configuration

korotkov-aerospike:
…gerBuilder interfaces; add unit tests

### DIFF
--- a/pkg/service/configuration_manager.go
+++ b/pkg/service/configuration_manager.go
@@ -16,13 +16,13 @@ type ConfigurationManager interface {
 	WriteConfiguration(config *model.Config) error
 }
 
-type Downloader interface {
-	Download(string) ([]byte, error)
+type Reader interface {
+	read(string) ([]byte, error)
 }
 
-type HTTPDownloader struct{}
+type HTTPReader struct{}
 
-func (h HTTPDownloader) Download(url string) ([]byte, error) {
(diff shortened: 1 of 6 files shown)


pkg/service/configuration_manager.go
```diff
@@ -55,27 +106,26 @@ func remoteConfigurationManager(configFile string, isDownload bool) (Configurati
...
+	if isDownload {
+		return b.http.Download(configFile)
+	} else {
+		return b.file.Download(configFile)
+	}
+}
 
-	buf := new(bytes.Buffer)
-	_, err = buf.ReadFrom(resp.Body)
+func isDownload(configFile string) (bool, error) {
```
reugn, line 124: `isHTTP` would be a better name.

pkg/service/configuration_manager.go
```diff
@@ -15,38 +16,88 @@ type ConfigurationManager interface {
 	WriteConfiguration(config *model.Config) error
 }
 
-func NewConfigurationManager(configFile string, remote bool) (ConfigurationManager, error) {
-	uri, err := url.Parse(configFile)
+type Downloader interface {
```
reugn, line 19: `Reader` with a `Read` method is preferable in this context, since you don't download a local file.

reugn (approved):

PR #189, opened 2024-06-18
APPS-1208 Fix s3 backups

korotkov-aerospike:


pkg/util/utils.go
```diff
@@ -41,3 +43,12 @@ func TryAndRecover(f func() string) (output string, err error) {
 	}()
 	return f(), err
 }
+
+func ParseS3Path(s string) (bucket string, path string, err error) {
```
reugn, line 47: Add a doc string explaining why `url.Parse(s)` didn't work in our case.

reugn (approved):

PR #191, opened 2024-07-07
APPS-1214 Close client after info command

korotkov-aerospike:


korotkov-aerospike:
That's main branch, with c client

reugn (approved):

PR #195, opened 2024-07-25
Add backup estimates endpoint

korotkov-aerospike:


pkg/model/current_backup.go
reugn, line 1: Document exported items.

pkg/model/current_backup.go
```diff
@@ -0,0 +1,65 @@
...
+	Full        *CurrentBackup `json:"full,omitempty"`
+	Incremental *CurrentBackup `json:"incremental,omitempty"`
+}
+
+type CurrentBackup struct {
+	TotalRecords     uint64     `json:"total_records,omitempty"`
+	DoneRecords      uint64     `json:"done_records,omitempty"`
+	StartTime        time.Time  `json:"start_time,omitempty"`
+	PercentageDone   int        `json:"percentage_done,omitempty"`
+	EstimatedEndTime *time.Time `json:"estimated_end_time,omitempty"`
```
reugn, line 19: We agreed to have a hyphen delimiter instead of the underscore.

pkg/service/backup_handler.go
```diff
@@ -101,6 +127,48 @@ func (h *BackupHandler) runFullBackupInternal(now time.Time) error {
...
+		}
+
+		h.fullBackupHandlers[namespace] = handler
+	}
+
+	return nil
+}
+
+func (h *BackupHandler) waitForFullBackups(ctx context.Context, now time.Time) error {
+	startTime := time.Now()
```
reugn, line 155: How does this `startTime` differ from the `now` parameter?

pkg/service/backup_handler.go
```diff
@@ -173,21 +232,32 @@ func (h *BackupHandler) runIncrementalBackup(now time.Time) {
...
+		slog.Error("failed to connect to aerospike cluster", "err", aerr)
 	}
+	defer func() {
+		client.Close()
+		clear(h.incrBackupHandlers)
+	}()
 
+	h.startIncrementalBackupForAllNamespaces(client, now)
+
+	h.waitForIncrementalBackups(context.TODO(), now)
```
reugn, line 252: Can the context be taken as a parameter?

internal/server/server.go
```diff
@@ -199,6 +205,8 @@ func (ws *HTTPServer) Start() {
 	// Schedules a full backup operation
 	mux.HandleFunc(ws.api("/backups/schedule/{name}"), ws.scheduleFullBackup)
 
+	mux.HandleFunc(ws.api("/backups/currentBackup/{name}"), ws.getCurrentBackupInfo)
```
reugn, line 208: Add a doc line.

internal/server/backup_handlers.go
```diff
@@ -198,3 +198,36 @@ func (ws *HTTPServer) scheduleFullBackup(w http.ResponseWriter, r *http.Request)
...
+	if routineName == "" {
+		http.Error(w, "routine name required", http.StatusBadRequest)
+		return
+	}
+
+	handler, found := ws.handlerHolder[routineName]
+	if !found {
+		http.Error(w, "unknown routine name "+routineName, http.StatusNotFound)
+		return
+
```
reugn, line 221: Empty line.

pkg/service/backup_handler.go
```diff
@@ -72,21 +85,34 @@ func (h *BackupHandler) runFullBackup(now time.Time) {
...
-		if err != nil {
-			return err
-		}
+
+	err = h.startFullBackupForAllNamespaces(now, client)
+	if err != nil {
+		return err
+	}
+
+	err = h.waitForFullBackups(context.TODO(), now)
```
reugn, line 113: Can the context be taken as a parameter?

pkg/service/backup_handler.go
```diff
@@ -173,21 +232,32 @@ func (h *BackupHandler) runIncrementalBackup(now time.Time) {
...
 	}
-	for _, namespace := range h.namespaces {
-		h.runIncrBackupForNamespace(now, namespace)
+	if len(h.incrBackupHandlers) > 0 {
+		slog.Debug("Incremental backup is currently in progress, skipping incremental backup",
+			"name", h.routineName)
+		return
+	}
+
+	client, aerr := aerospike.NewClientWithPolicyAndHost(h.cluster.ASClientPolicy(), h.cluster.ASClientHosts()...)
```
reugn, line 241: Is it possible to reuse the client for full and incremental backups in the same routine?

pkg/service/backup_handler.go
korotkov-aerospike: startTime is used only to calculate duration.
now — is a execution time from scheduler, we use it in business logic.
They might be almost equal, but still the meaning is different.

pkg/service/backup_handler.go
korotkov-aerospike: maybe, but its another ticket.
that's a general question, do we want to have client idle connected between backups? I'm not sure.

reugn (approved):

pkg/service/backup_handler.go
reugn: Yes, let's address this in a separate discussion.

PR #199, opened 2024-07-28
APPS-1228 Don't create multiple clients on every backup/restore

korotkov-aerospike:
Partly this ticket was done with other refactoring. Now I reuse client for configuration backup.
Now, we create client totally for 5 reasons:
1-2 every full/incr backup
3-4 every restore (by path or by timestamp)
5) once on application startup to fetch namespaces to backup.


korotkov-aerospike:
http://3.65.168.247:8080/ui/#pitr/launches/all/7272
http://3.65.168.247:8080/ui/#pitr/launches/all/7270
all tests pass

pkg/service/restore_memory.go
```diff
@@ -69,29 +83,51 @@ func (r *RestoreMemory) RestoreByTime(request *model.RestoreTimestampRequest) (i
...
 	request *model.RestoreTimestampRequest,
 	jobID int,
 	fullBackups []model.BackupDetails,
 ) {
+	client, aerr := aerospike.NewClientWithPolicyAndHost(
+		request.DestinationCuster.ASClientPolicy(),
+		request.DestinationCuster.ASClientHosts()...)
+	if aerr != nil {
+		err := fmt.Errorf("failed to connect to aerospike cluster, %w", aerr)
+		slog.Error("Failed to restore by timestamp", "routine", request.Routine, "err", err)
```
reugn, line 98: The log entry is different from the one written in `Restore`. Is it possible to extract this client initialization block into a method?

korotkov-aerospike: good point.

PR #213, opened 2024-08-13
APPS-1270 Fetch namespaces on backup

korotkov-aerospike:


korotkov-aerospike:
If no namespace provided in config, we need to backup all => make info call on every backup.
If new NS would be added during service run, it will be backed up too.

reugn (approved):

PR #214, opened 2024-08-13
APPS-1252 Optimize validation for S3

korotkov-aerospike:
Apply took long, because for every routine we checked s3 connection.
1) Don't fail if bucket (or S3 in general) is not available on startup. We will fail on actual backup anyway.
2) Move parsing of path into validation, to simplify process.

### DIFF
--- a/pkg/service/s3_context.go
+++ b/pkg/service/s3_context.go
@@ -77,7 +77,9 @@ func checkBucket(ctx context.Context, client *s3.Client, bucket string) {
 	})
 
 	if err != nil {
-		slog.Warn("AWS S3 Bucket don't exist", slog.String("bucket", bucket), "err", err)
+		slog.Warn("AWS S3 Bucket don't exist",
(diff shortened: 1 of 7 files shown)


pkg/service/s3_context.go
```diff
@@ -51,39 +47,55 @@ func NewS3Context(storage *model.Storage) (*S3Context, error) {
...
 	storage.SetDefaultProfile()
-	return config.LoadDefaultConfig(
+	cfg, err := config.LoadDefaultConfig(
 		ctx,
 		config.WithSharedConfigProfile(*storage.S3Profile),
 		config.WithRegion(*storage.S3Region),
 	)
+
+	if err != nil {
+		panic(fmt.Sprintf("failed loading config, %v", err))
```
reugn, line 95: Add a TODO to revisit this panic.

reugn (approved):

PR #221, opened 2024-08-14
Reuse Aerospike client in fetch namespaces request

korotkov-aerospike:


korotkov-aerospike:
Also remove RemoveArtifacts flag, it's not used.

korotkov-aerospike:
Now all Aerospike clients are created (and cached) using client manager.

reugn (approved):

PR #223, opened 2024-08-18
APPS-1275 Implement retry logic in restore operation

korotkov-aerospike:


pkg/model/retry_policy.go
```diff
@@ -0,0 +1,47 @@
...
+	MaxRetries int `json:"maxRetries" yaml:"maxRetries"`
+}
+
+// Validate checks if the RetryPolicy fields are valid.
+func (rp *RetryPolicy) Validate() error {
+	if rp == nil {
+		return nil
+	}
+
+	if rp.BaseTimeout <= 0 {
```
filkeith, line 29: Why not uint64? if it must be positive

pkg/model/retry_policy.go
```diff
@@ -0,0 +1,47 @@
...
+		return fmt.Errorf("BaseTimeout must be greater than 0")
+	}
+
+	if rp.Multiplier < 1 {
+		return fmt.Errorf("multiplier must be greater or equal than 1")
+	}
+
+	if rp.MaxRetries < 0 {
```
filkeith, line 37: same here uint?

filkeith (commented):
do we still need `modules/schema` ? and modules at all

pkg/model/retry_policy.go
korotkov-aerospike: This is DTO object.
What if user passes -1? It will be decoded into a really big int.

korotkov-aerospike:
@filkeith we do need schema, it is used to fetch aerospike configuration.

reugn (approved):

PR #236, opened 2024-10-06
APPS-1279 Support GCP storage

korotkov-aerospike:


pkg/service/storage/accessor.go
```diff
@@ -0,0 +1,36 @@
...
+// Accessor interface abstracts storage layer.
+type Accessor interface {
+	supports(storage model.Storage) bool
+	createReader(ctx context.Context, storage model.Storage, path string, isFile bool, filter Validator,
+	) (backup.StreamingReader, error)
+	createWriter(ctx context.Context, storage model.Storage, path string, isFile, isRemoveFiles, withNested bool,
+	) (backup.Writer, error)
+}
+
+var accessors []Accessor
```
reugn, line 20: I believe you have considered another option than a global variable.

pkg/service/storage/accessor.go
```diff
@@ -0,0 +1,36 @@
...
+
+// registerAccessor adds a new storage accessor.
+func registerAccessor(accessor Accessor) {
+	accessors = append(accessors, accessor)
+}
+
+// getAccessor returns the appropriate accessor for the given storage.
+func getAccessor(storage model.Storage) Accessor {
+	for _, accessor := range accessors {
```
reugn, line 29: It would be better to make `accessors` a map for performance reasons even though it is negligible.

pkg/service/storage/validator.go
```diff
@@ -0,0 +1,34 @@
+package storage
+
+import (
+	"fmt"
+	"strings"
+)
+
+type Validator interface {
```
reugn, line 9: Add a doc line explaining the need for this validator.

pkg/service/storage/accessor.go
korotkov-aerospike: yes, I just though it's easier

reugn (approved):

filkeith (approved):
lgtm

PR #239, opened 2024-10-08
APPS-1299 Add Azure Blob Storage support

korotkov-aerospike:


pkg/model/storage.go
```diff
@@ -1,49 +1,95 @@
...
+	Auth AzureAuth
+}
+
+func (s *AzureStorage) storage() {}
+
+// AzureAuth represents the authentication methods for Azure Blob storage.
+// This interface is implemented by AzureSharedKeyAuth and AzureADAuth.
+type AzureAuth interface {
+	azureAuth()
+}
```
reugn, line 87: Consider using the constraint interface:
```go
type AzureAuth interface {
	AzureSharedKeyAuth | AzureADAuth
}
```
but this will require additional changes in the current logic.

korotkov-aerospike: no, it's not allowed, I tried.
you cannot have a field of type AzureAuth in this case

reugn:
```go
type AzureStorage[T AzureAuth] struct {
	// Path is the root directory within the Azure Blob container where backups will be stored.
	Path string
	// Endpoint is the URL of the Azure Blob storage service.
	Endpoint string
	// ContainerName is the name of the Azure Blob container where backups will be stored.
	ContainerName string
	// Auth holds the authentication details for Azure Blob storage.
	// It can be either AzureSharedKeyAuth or AzureADAuth.
	Auth T
}
```
But more changes will be required.

korotkov-aerospike: it seems way too many for the benefit we get.

reugn: Can be revisited later.

reugn (approved):

filkeith (approved):
lgtm

PR #241, opened 2024-10-09
Performance optimization: use StartAfter flag in file list scans

korotkov-aerospike:
StartAfter flag was implemented in FMWK-559 in lib.

### DIFF
--- a/pkg/service/storage/azure.go
+++ b/pkg/service/storage/azure.go
@@ -20,7 +20,7 @@ func (a *AzureStorageAccessor) supports(storage model.Storage) bool {
 }
 
 func (a *AzureStorageAccessor) createReader(
-	ctx context.Context, storage model.Storage, path string, isFile bool, filter Validator,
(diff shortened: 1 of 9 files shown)


pkg/service/storage/azure.go
```diff
@@ -20,7 +20,7 @@ func (a *AzureStorageAccessor) supports(storage model.Storage) bool {
 }
 
 func (a *AzureStorageAccessor) createReader(
-	ctx context.Context, storage model.Storage, path string, isFile bool, filter Validator,
+	ctx context.Context, storage model.Storage, path string, isFile bool, filter Validator, from string,
```
filkeith: may be name it `startFrom` ?  (and in all other places)

korotkov-aerospike: maybeEven startScanFrom ?

filkeith: yep it would be great, because `from` is to abstract

filkeith (approved):
lgtm

PR #242, opened 2024-10-09
Ignore S3 endpoint override field when it's an empty string

korotkov-aerospike:
This check was unfortunately lost during refactoring.

### DIFF
--- a/pkg/service/storage/s3.go
+++ b/pkg/service/storage/s3.go
@@ -79,7 +79,7 @@ func getS3Client(ctx context.Context, s *model.S3Storage) (*awsS3.Client, error)
 	}
 
 	client := awsS3.NewFromConfig(cfg, func(o *awsS3.Options) {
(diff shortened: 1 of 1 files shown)


pkg/service/storage/s3.go
```diff
@@ -79,7 +79,7 @@ func getS3Client(ctx context.Context, s *model.S3Storage) (*awsS3.Client, error)
 	}
 
 	client := awsS3.NewFromConfig(cfg, func(o *awsS3.Options) {
-		if s.S3EndpointOverride != nil {
+		if s.S3EndpointOverride != nil && *s.S3EndpointOverride != "" {
```
reugn, line 82: Do we have a test for this?

korotkov-aerospike: it's only valid for AWS installation.
so, yes, or K8 script is a test

reugn (approved):

PR #244, opened 2024-10-23
APPS-1268 Validate backup/restore namespace exists

korotkov-aerospike:


pkg/service/backup_go.go
```diff
@@ -33,6 +33,14 @@ func (b *BackupGo) BackupRun(
 	namespace string,
 	path string,
 ) (BackupHandler, error) {
+	ok, err := clusterHasRequiredNamespace(namespace, client.AerospikeClient())
+	if err != nil {
+		return nil, err
+	}
+	if !ok {
+		return nil, fmt.Errorf("cluster does not have all require namespace")
```
reugn: > return nil, fmt.Errorf("cluster does not have the namespace %s configured", namespace)

pkg/service/backup_go.go
```diff
@@ -33,6 +33,14 @@ func (b *BackupGo) BackupRun(
 	namespace string,
 	path string,
 ) (BackupHandler, error) {
+	ok, err := clusterHasRequiredNamespace(namespace, client.AerospikeClient())
```
reugn: Won't it be better to validate the namespace either at startup or when modified using an API endpoint?

pkg/service/aerospike_service.go
```diff
@@ -16,6 +16,22 @@ import (
 
 const namespaceInfo = "namespaces"
 
+// clusterHasRequiredNamespace checks if given namespace exists in cluster.
+func clusterHasRequiredNamespace(namespace string, client backup.AerospikeClient) (bool, error) {
```
reugn: Did you consider verifying the namespace on a restore operation?

korotkov-aerospike:
I've introduced nsValidator component that does the info request and validates namespaces.
If aerospike is unavailable, validation is skipped.
I do it when the config is updated, or routine or cluster. It had added a bit of complexity, but I think it's worth it.

pkg/service/aerospike_service.go
```diff
@@ -16,6 +18,75 @@ import (
...
+	}
+
+	return util.MissingElements(namespaces, namespacesInCluster)
+}
+
+func (nv *defaultNamespaceValidator) ValidateRoutines(cluster *model.AerospikeCluster, config *model.Config) error {
+	var err error
+	routines := filterRoutinesByCluster(config.BackupRoutines, cluster)
+	for routineName, routine := range routines {
+		missingNamespaces := nv.MissingNamespaces(cluster, routine.Namespaces)
```
reugn, line 67: This will create a new client each time, even though the routines are likely to share the cluster.

pkg/service/restore_data.go
```diff
@@ -89,6 +96,25 @@ func (r *dataRestorer) Restore(request *model.RestoreRequest) (model.RestoreJobI
...
+// validateDestinationNamespace checks if destination cluster contains namespace from restore request (if it is set).
+func (r *dataRestorer) validateDestinationNamespace(request *model.RestoreRequest, jobID model.RestoreJobID) bool {
+	if request.Policy.Namespace != nil {
+		missingNamespaces := r.nsValidator.MissingNamespaces(
+			request.DestinationCuster, []string{*request.Policy.Namespace.Destination})
+		if len(missingNamespaces) > 0 {
+			// it can be only 1 missing ns.
+			err := fmt.Errorf("destination cluster does not have namespace %s", missingNamespaces[0])
+			slog.Error("Failed to restore by path",
+				slog.Any("cluster", request.DestinationCuster),
```
reugn: This logging may leak credentials.

pkg/service/aerospike_service.go
korotkov-aerospike: no, we use nv.ClientManager.GetClient inside, so client will be taken from cache.

pkg/service/restore_data.go
reugn: * `request.DestinationCuster` is a pointer.
* DestinationCuster -> DestinationCluster

pkg/service/aerospike_service.go
reugn: It is being closed and removed from the cache right after the use.

reugn (approved):

PR #247, opened 2024-11-05
APPS-1325 Delete failed backups

korotkov-aerospike:
PR is fixing unstable integration tests after introducing client cache
1) Delete failed full backup (per namespace)
2) Check connection status before getting client from cache

### DIFF
--- a/pkg/service/client_manager.go
+++ b/pkg/service/client_manager.go
@@ -111,7 +111,7 @@ func isClusterHealthy(client backup.AerospikeClient) bool {
 		return false
 	}
 
-	info, err := node.RequestInfo(nil, "status")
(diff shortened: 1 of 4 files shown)


pkg/service/client_manager.go
```diff
@@ -73,17 +99,21 @@ func (cm *ClientManagerImpl) GetClient(cluster *model.AerospikeCluster) (*backup
 }
 
 // getExistingClient tries to get an existing client from the cache.
-// Returns nil if client doesn't exist.
-func (cm *ClientManagerImpl) getExistingClient(cluster *model.AerospikeCluster) *backup.Client {
+// Returns nil if client doesn't exist, error is client is not connected.
```
reugn: is -> if

reugn (approved):

PR #250, opened 2024-11-06
Update dependencies versions

korotkov-aerospike:


korotkov-aerospike:
We need to update lib version, and the rest are just at the same time

reugn (approved):

PR #255, opened 2024-11-13
APPS-1335 Don't save namespace handler if it fails to start

korotkov-aerospike:
The actual error was a permission to AS cluster

time=2024-11-13T13:24:00.236Z level=INFO source=:0 msg="ERROR error reading record backup.client.id=asd.aerospike.com backup.handler.id=2cbaae67-c10b-47ba-8d31-fa6ca1b133dc backup.handler.type=backup backup.handler.storage=s3 backup.reader.id=e52efeb8-cb20-4220-911c-af2ba5ed2550 backup.reader.type=record backup.error=\"ResultCode: **FAIL_FORBIDDEN**, Iteration: 0, InDoubt: false, Node: BC40300000A0142 10.0.0.3:4333: Operation not allowed at this time\\n  ResultCode: FAIL_FORBIDDEN, Iteration: 0, InDoubt: false, Node: <nil>: Operation not allowed at this time\""

So, backup failed to start and returned error. But unfortunately, there was no _continue_ in error handling, so failed (nil) handler was added to the map and caused crash. Thats an old bug, we didn't got it before because we had no connection issues, in our negative tests we only tested no connection to databased, but not access denied.

### DIFF
--- a/pkg/service/retry_service.go
+++ b/pkg/service/retry_service.go
@@ -8,20 +8,30 @@ import (
 	"time"
 )
 
-// RetryService is a service for retrying a function with a specified interval
+type Executor interface {
+	run(label string, f func() error) error
+}
+
+type SimpleExecutor struct{}
+
+func (e *SimpleExecutor) run(_ string, f func() error) error {
+	return f()
+}
+
+// RetryExecutor is a service for retrying a function with a specified interval
 // and a maximum number of attempts.
-type RetryService struct {
+type RetryExecutor struct {
 	logger        *slog.Logger
 	retryInterval time.Duration
 	maxAttempts   int
 }
 
-// NewRetryService returns a new RetryService instance.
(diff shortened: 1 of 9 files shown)


pkg/service/retry_service.go
```diff
@@ -8,20 +8,30 @@ import (
 	"time"
 )
 
-// RetryService is a service for retrying a function with a specified interval
+type Executor interface {
```
reugn: * Add a doc, explaining the label part.
* Can be unexported.

pkg/service/retry_service.go
```diff
@@ -8,20 +8,30 @@ import (
...
+	run(label string, f func() error) error
+}
+
+type SimpleExecutor struct{}
```
reugn: Can be unexported.

reugn (approved):

PR #259, opened 2024-11-17
APPS-1338 Fix nil dereference on backup reading

korotkov-aerospike:


korotkov-aerospike:
I had updated lib version and added few more logs.
All tests on Jenkins passed.

reugn (approved):

PR #264, opened 2024-11-20
APPS-1351 Last backup run field optimization

korotkov-aerospike:
refactoring 
1) rename BackupState to LastRun and simplify it.

performance optimisation 
2) scan backups for last run in parallel
3) try to reuse last backup run property after config reload


korotkov-aerospike:
1) rename BackupState to LastRun and simplify it.
2) scan backups for last run in parallel
3) try to reuse last backup run property after config reload

reugn (approved):

PR #267, opened 2024-11-24
APPS-1354 Relax storage configuration validation for GCP and Azure

korotkov-aerospike:


korotkov-aerospike:
@davi17g confirmed that GCP works.

reugn (approved):

PR #277, opened 2024-11-28
APPS-1365 Expose config and requests validation for external use

korotkov-aerospike:
I created package validation with all functions AKO team will need. It's always possible to add more on demand.
Many files changed because I moved comparison logic from model to dto package

### DIFF
--- a/pkg/validation/config_validation.go
+++ b/pkg/validation/config_validation.go
@@ -10,10 +10,6 @@ import (
 
 // ConfigurationApplicable checks if new config changes are allowed as per dynamic consideration.
 func ConfigurationApplicable(oldConf, newConf *dto.Config) error {
-	if err := ValidateConfiguration(newConf); err != nil {
-		return err
-	}
-
 	// currently, only ServiceConfig can not be changed dynamically.
 	return oldConf.ServiceConfig.Compare(newConf.ServiceConfig)
 }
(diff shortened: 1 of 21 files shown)


pkg/validation/config_validation.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func ValidateConfiguration(conf *dto.Config) error {
+	clientManager := service.NewClientManager(&service.DefaultClientFactory{}, time.Second)
+	namespaceValidator := service.NewNamespaceValidator(clientManager)
+	_, err := conf.ToModel(namespaceValidator)
+	return err
+}
+
+func ValidateRequest(request dto.RestoreRequest, conf *dto.Config) error {
```
abhishekdwivedi3060: nit: `ValidateRestoreRequest`
`ValidateRequest` name is very generic but this func is only related to restore request

pkg/validation/config_validation.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/aerospike/aerospike-backup-service/v2/pkg/dto"
+	"github.com/aerospike/aerospike-backup-service/v2/pkg/service"
+)
+
+// ConfigurationApplicable checks if new config changes are allowed as per dynamic consideration.
+func ConfigurationApplicable(oldConf, newConf *dto.Config) error {
```
abhishekdwivedi3060: How about something like `CompareConfigsForStaticChanges`, `ValidateStaticFieldChanges` or `CompareStaticFields`?

abhishekdwivedi3060 (approved):

PR #279, opened 2024-12-04
APPS-1378 Make `dto` pkg compatible for external use

korotkov-aerospike:


korotkov-aerospike:
Regression tests passed during daily

abhishekdwivedi3060 (approved):

PR #286, opened 2024-12-19
APPS-1412 Default values for server and logger

korotkov-aerospike:


korotkov-aerospike:
Also, use time.Duration instead of int for time-related models fields.

internal/server/handlers/config_policy_test.go
```diff
@@ -18,7 +18,7 @@ const (
 )
 
 func testConfigBackupPolicy() *dto.BackupPolicy {
-	testIn32 := int32(10)
+	testIn32 := 10
```
reugn: Can be removed and `testInt` be used instead.

reugn (approved):

PR #289, opened 2024-12-23
Automate README example generation

korotkov-aerospike:
I want to verify that the examples are valid and up to date


cmd/readme/readme_example_generator.go
```diff
@@ -0,0 +1,196 @@
+package main
```
reugn: Relocate the script under `build/readme`.

reugn (commented):
Great initiative to automate example generation.

reugn (approved):

PR #291, opened 2024-12-24
APPS-1417 Respect scheduled full backup timing

korotkov-aerospike:


korotkov-aerospike:
In scheduleFullBackup I removed running backup now.
Also, extracted interface for *BackupRoutineHandler and with it was able to add unit tests.
2 minor bugs found
1) We should not add incremental backups to jobStore
2) We should clean jobStore on config reload. jobStore should not be global variable in the first place, but it's too much of a change to fix it now.

Also, I added last run to CurrentBackup stats.

pkg/model/last_run.go
```diff
@@ -0,0 +1,30 @@
...
+		Full:        lastFullBackup,
+		Incremental: lastIncrBackup,
+	}
+}
+
+func (r *LastBackupRun) NoFullBackup() bool {
+	return r.Full == nil
+}
+
+func (r *LastBackupRun) LastAnyRun() *time.Time {
```
reugn: Consider renaming to `LatestRun`.

pkg/model/last_run.go
reugn, line 1: last_backup_run.go

pkg/service/backup_routine_handler.go
```diff
@@ -158,7 +172,7 @@ func (h *BackupRoutineHandler) runFullBackupInternal(ctx context.Context, now ti
 		return err
 	}
 
-	h.lastRun.full = now
+	h.lastRun.Full = &now
```
reugn: It is not a new thing, but it is better to synchronize writes and reads to prevent data races.

pkg/service/backup_routine_handler.go
```diff
@@ -328,7 +342,7 @@ func (h *BackupRoutineHandler) runIncrementalBackupInternal(ctx context.Context,
 		return err
 	}
 
-	h.lastRun.incremental = now
+	h.lastRun.Incremental = &now
```
reugn: Consider synchronizing.

pkg/model/current_backup.go
```diff
@@ -10,6 +10,8 @@ type CurrentBackups struct {
 	Full *RunningJob
 	// Incremental represents the state of an incremental backup. Nil if no incremental backup is running.
 	Incremental *RunningJob
+	// LastRunTime: the last time when a backup was run
```
reugn: > LastRunTime contains information about the latest run time for both full and incremental backups.

pkg/model/last_run.go
```diff
@@ -0,0 +1,30 @@
...
+
+// LastBackupRun stores the last run times for both full and incremental backups.
+type LastBackupRun struct {
+	// Last time the Full backup was performed.
+	Full *time.Time
+	// Last time the Incremental backup was performed.
+	Incremental *time.Time
+}
+
+func NewLastRun(lastFullBackup *time.Time, lastIncrBackup *time.Time) LastBackupRun {
```
reugn: Consider renaming to `MakeLastBackupRun` or `CreateLastBackupRun`, since it is not a good practice to return a value for `New` constructors.

pkg/service/backup_routine_handler.go
korotkov-aerospike: No backup of the same type for the same routine will run simultaneously, no need for lock

pkg/service/backup_routine_handler.go
korotkov-aerospike: No backup of the same type for the same routine will run simultaneously, no need for lock

reugn: There are two jobs scheduled for the full and incremental backup that may concurrently access those variables.

pkg/model/last_backup_run.go
```diff
@@ -0,0 +1,50 @@
...
+	"time"
+)
+
+// LastBackupRun stores the last run times for both full and incremental backups.
+type LastBackupRun struct {
+	mu sync.RWMutex
+	// Last time the Full backup was performed.
+	Full *time.Time
+	// Last time the Incremental backup was performed.
+	Incremental *time.Time
```
reugn: Make the fields private to prevent setting them directly. Add getters if required.

pkg/model/last_backup_run.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func NewLastRun(lastFullBackup *time.Time, lastIncrBackup *time.Time) *LastBackupRun {
```
reugn: NewLastBackupRun

reugn (approved):

PR #294, opened 2024-12-29
APPS-1433 Fix Docker build for arm64 in v2

davi17g:


.github/workflows/image-release.yaml
```diff
@@ -1,77 +1,77 @@
-name: Build Release Docker Images
-on:
-  push:
-    tags:
-      - 'v*'
```
reugn: Do not comment it out. Replace the tag trigger with `workflow_dispatch`.

reugn (approved):

PR #295, opened 2024-12-29
APPS-1431 Synchronize access to the service configuration

korotkov-aerospike:


korotkov-aerospike:
1) moved lock from config to service level. 
2) reduced number of pointers to config. Actually, now after initial startup, it is only stored in http service. So, every request to config goes through http service. 
Periodic backups dont access config.

pkg/model/config.go
```diff
@@ -7,17 +7,53 @@ import (
...
+
+	newConfig := &BackupConfig{
+		AerospikeClusters: make(map[string]*AerospikeCluster, len(bc.AerospikeClusters)),
+		Storage:           make(map[string]Storage, len(bc.Storage)),
+		BackupPolicies:    make(map[string]*BackupPolicy, len(bc.BackupPolicies)),
+		BackupRoutines:    make(map[string]*BackupRoutine, len(bc.BackupRoutines)),
+		SecretAgents:      make(map[string]*SecretAgent, len(bc.SecretAgents)),
+	}
+
+	for k, v := range bc.AerospikeClusters {
```
reugn: Consider using
```go
maps.Copy(newConfig.AerospikeClusters, bc.AerospikeClusters)
```

pkg/model/config.go
```diff
@@ -7,17 +7,53 @@ import (
 
 // Config represents the service configuration.
 type Config struct {
-	mu                sync.Mutex
-	ServiceConfig     BackupServiceConfig
+	sync.RWMutex
```
reugn: As the plan is to use the mutex internally, embedding can be reverted.

pkg/model/config.go
```diff
@@ -26,60 +62,72 @@ func NewConfig() *Config {
...
+	}
+}
+
 var (
 	ErrAlreadyExists = fmt.Errorf("item already exists")
 	ErrNotFound      = fmt.Errorf("item not found")
 	ErrInUse         = fmt.Errorf("item is in use")
 )
 
+func (c *Config) GetBackupConfig() *BackupConfig {
```
reugn: Consider renaming to `CopyBackupConfig`.

reugn (approved):

PR #296, opened 2024-12-30
Update documentation for v3 new features

korotkov-aerospike:
* Backup Retention Policy
* Cancel Restore Job
* Cancel Backup Job
* Disable Routine

### DIFF
--- a/README.md
+++ b/README.md
@@ -666,7 +666,7 @@ The feature allows users to specify retention rules for both full and incrementa
 Retention policy is (optional) part of Backup Policy. It consists of 2 integer fields.
 
 * `full`: The total number of full backups to retain. If absent, all full backups are kept. Minimum is 1, it means that each new full backup will delete the previous one.
-* `incremental`: The number of full backups to retain along with their incremental backups. Cannot exceed `full`. Minimum is 0, it means that all old incremental backups will be deleted after each full. 
+* `incremental`: The number of full backups to retain their incremental backups. Cannot exceed `full`. Minimum is 0, it means that all old incremental backups will be deleted after each full. 
 
 If no RetentionPolicy is specified, the system will default to retaining all backups, mimicking the previous KeepAll behavior.
 
--- a/README.md
+++ b/README.md
@@ -663,7 +663,7 @@ was changed to simplify and streamline the process.
 
 This release introduces a new, configurable **RetentionPolicy** for managing backup storage more effectively.
 The feature allows users to specify retention rules for both full and incremental backups, replacing the previous options `KeepAll`, `RemoveAll`, and `RemoveIncremental`.
-Retention policy is (optional) part of Backup Policy. It consists of 2 integer fields.
+Retention policy is an optional part of a backup policy. It consists of two integer fields:
 
 * `full`: The total number of full backups to retain. If not specified, all full backups are kept. The minimum is 1, meaning each new full backup deletes the previous one.
 * `incremental`: The number of full backups to retain their incremental backups. Cannot exceed `full`. Minimum is 0, it means that all old incremental backups will be deleted after each full. 
--- a/README.md
+++ b/README.md
@@ -662,7 +662,7 @@ was changed to simplify and streamline the process.
 ### Backup Retention Policy
 
 This release introduces a new, configurable **RetentionPolicy** for managing backup storage more effectively.
-The feature allows users to specify retention rules for both full and incremental backups, replacing the previous rigid options (`KeepAll`, `RemoveAll`, `RemoveIncremental`).
+The feature allows users to specify retention rules for both full and incremental backups, replacing the previous options `KeepAll`, `RemoveAll`, and `RemoveIncremental`.
 Retention policy is (optional) part of Backup Policy. It consists of 2 integer fields.
 
 * `full`: The total number of full backups to retain. If absent, all full backups are kept. Minimum is 1, it means that each new full backup will delete the previous one.
--- a/README.md
+++ b/README.md
@@ -671,7 +671,7 @@ Retention policy is (optional) part of Backup Policy. It consists of 2 integer f
 If no retention policy is specified, the system defaults to retaining all full and incremental backups, the same as the `KeepAll` value in older versions.
 
 After each successfull full backup, all existing backups are scanned to count full and incremental backups.
-Service then removes older full backups and their associated incrementals as needed to retain only
+ABS then removes older full backups and their associated incremental backups as needed to retain only
 the last `full` backups and incremental backups for the most recent `incemental` backups.
 
 ## New API functions (v2 → v3):
--- a/README.md
+++ b/README.md
@@ -672,7 +672,7 @@ If no retention policy is specified, the system defaults to retaining all full a
 
 After each successfull full backup, all existing backups are scanned to count full and incremental backups.
 ABS then removes older full backups and their associated incremental backups as needed to retain only
-the last `full` backups and incremental backups for the most recent `incemental` backups.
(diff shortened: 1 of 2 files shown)


reugn (approved):

README.md
```diff
@@ -652,6 +659,21 @@ was changed to simplify and streamline the process.
   The specific backup data location is now specified using a new required field: **`backup-data-path`**.
   This change allows you to reuse the same storage for different restore requests.
 
+### Backup Retention Policy
+
+This release introduces a new, configurable **RetentionPolicy** for managing backup storage more effectively.
+The feature allows users to specify retention rules for both full and incremental backups, replacing the previous rigid options (`KeepAll`, `RemoveAll`, `RemoveIncremental`).
```
alex-thomas-aerospike:
```suggestion
The feature allows users to specify retention rules for both full and incremental backups, replacing the previous options `KeepAll`, `RemoveAll`, and `RemoveIncremental`.
```

README.md
```diff
@@ -652,6 +659,21 @@ was changed to simplify and streamline the process.
...
+Retention policy is (optional) part of Backup Policy. It consists of 2 integer fields.
```
alex-thomas-aerospike:
```suggestion
Retention policy is an optional part of a backup policy. It consists of two integer fields:
```

README.md
```diff
@@ -652,6 +659,21 @@ was changed to simplify and streamline the process.
...
+
+* `full`: The total number of full backups to retain. If absent, all full backups are kept. Minimum is 1, it means that each new full backup will delete the previous one.
```
alex-thomas-aerospike:
```suggestion
* `full`: The total number of full backups to retain. If not specified, all full backups are kept. The minimum is 1, meaning each new full backup deletes the previous one.
```

README.md
```diff
@@ -652,6 +659,21 @@ was changed to simplify and streamline the process.
...
+* `incremental`: The number of full backups to retain their incremental backups. Cannot exceed `full`. Minimum is 0, it means that all old incremental backups will be deleted after each full.
```
alex-thomas-aerospike: I'm not clear exactly on how this parameter works, can you give more examples in a comment of what would happen with a value of 1, or what happens when it's the same as `full`? What happens if the user enters a value equal to the value of `full`?

README.md
```diff
@@ -652,6 +659,21 @@ was changed to simplify and streamline the process.
...
+### Backup Retention Policy
+
+This release introduces a new, configurable **RetentionPolicy** for managing backup storage more effectively.
+The feature allows users to specify retention rules for both full and incremental backups, replacing the previous rigid options (`KeepAll`, `RemoveAll`, `RemoveIncremental`).
+Retention policy is (optional) part of Backup Policy. It consists of 2 integer fields.
+
+* `full`: The total number of full backups to retain. If absent, all full backups are kept. Minimum is 1, it means that each new full backup will delete the previous one.
+* `incremental`: The number of full backups to retain their incremental backups. Cannot exceed `full`. Minimum is 0, it means that all old incremental backups will be deleted after each full. 
+
+If no RetentionPolicy is specified, the system will default to retaining all backups, mimicking the previous KeepAll behavior.
```
alex-thomas-aerospike:
```suggestion
If no retention policy is specified, the system defaults to retaining all full and incremental backups, the same as the `KeepAll` value in older versions.
```

README.md
```diff
@@ -652,6 +659,21 @@ was changed to simplify and streamline the process.
...
+
+After each successfull full backup, all existing backups are scanned to count full and incremental backups.
+Service then removes older full backups and their associated incrementals as needed to retain only
```
alex-thomas-aerospike:
```suggestion
ABS then removes older full backups and their associated incremental backups as needed to retain only
```

README.md
```diff
@@ -652,6 +659,21 @@ was changed to simplify and streamline the process.
...
+the last `full` backups and incremental backups for the most recent `incemental` backups.
```
alex-thomas-aerospike:
```suggestion
the last `full` backups and incremental backups for the most recent `incremental` backups.
```

README.md
```diff
@@ -722,3 +744,29 @@ Validation allows only one of these three fields to be present.
...
+
+
+### Disable Routine
+
+New endpoints:
+
+- [`POST {{baseUrl}}/v1/routines/:<routineName>/disable/`](https://aerospike.github.io/aerospike-backup-service/#/Routine/disable)
+- [`POST {{baseUrl}}/v1/routines/:<routineName>/enable/`](https://aerospike.github.io/aerospike-backup-service/#/Routine/enable)
+
+Set the disabled flag for the given routine to `true` or `false` (default is `false`).
```
alex-thomas-aerospike, line 769: These github.io links don't go to the correct section on the live page. Can you update them, or will the openAPI spec get updated to reflect this?

alex-thomas-aerospike (changes requested):

README.md
korotkov-aerospike: Yes, the API is not published yet.

README.md
korotkov-aerospike: nil means no deletion.
full = nil, incr = nil => keep all
full = nil, incr = 0 => keep all full, delete all incremental after next full backup. So, incremental backups will actually happen, but they will live only up to next full backup.
full = 0 => forbidden
full = 1 => only store one full backup
full = N, incr = M => store N full backups, for last M of them store also incremental backups. N >= M.
full = N, incr = N => store N full backups, all with incrementals.

Does that make sense?

alex-thomas-aerospike: Got it, I think I would rewrite it as such, please check if this is right:

`* `incremental`: The number of most recent full backups that also retain intermediate backups made between them. Cannot exceed the value of `full`. If omitted, all incremental backups are kept. A value of 0 means that all previous existing incremental backups will be deleted after each full backup is made.`

Please also add one or two lines saying what happens with the forbidden cases of "full = 0" and "incremental > full". Error? Fall back to default of nil?

korotkov-aerospike: We never used word "intermediate" before, let's stick with "incremental".
In forbidden cases it will fail validation on startup. Just like any other field, we have a lot of validation rules.

alex-thomas-aerospike: Yes, you're right of course on "incremental." I'll add something in the main docs about validation checks, no need to add it here. Copy in that wording and I'll re-review, thanks!

README.md
```diff
@@ -663,16 +663,16 @@ was changed to simplify and streamline the process.
 
 This release introduces a new, configurable **RetentionPolicy** for managing backup storage more effectively.
 The feature allows users to specify retention rules for both full and incremental backups, replacing the previous options `KeepAll`, `RemoveAll`, and `RemoveIncremental`.
-Retention policy is (optional) part of Backup Policy. It consists of 2 integer fields.
+Retention policy is an optional part of a backup policy. It consists of two integer fields:
 
-* `full`: The total number of full backups to retain. If absent, all full backups are kept. Minimum is 1, it means that each new full backup will delete the previous one.
+* `full`: The total number of full backups to retain. If not specified, all full backups are kept. The minimum is 1, meaning each new full backup deletes the previous one.
 * `incremental`: The number of full backups to retain their incremental backups. Cannot exceed `full`. Minimum is 0, it means that all old incremental backups will be deleted after each full.
```
alex-thomas-aerospike:
```suggestion
* `incremental`: The number of most recent full backups that also retain incremental backups made between them. Cannot exceed the value of `full`. If omitted, all incremental backups are kept. A value of `0` means that all previous existing incremental backups will be deleted after each full backup is made.
```

alex-thomas-aerospike (approved):

PR #299, opened 2025-01-01
APPS-1436 Add static fields validation in update config handler

korotkov-aerospike:


pkg/service/config_applier.go
```diff
@@ -18,6 +18,7 @@ type ConfigApplier interface {
 }
 
 type DefaultConfigApplier struct {
+	mu            sync.RWMutex
```
reugn: Could use a regular mutex.

reugn (approved):

PR #300, opened 2025-01-06
APPS-1440 Synchronise access to metadata files and backup job

korotkov-aerospike:


korotkov-aerospike:
1) Read/Write metadata is synched
2) jobStore is locked as a whole, not for individual operations.

reugn (approved):

PR #302, opened 2025-01-09
Address data synchronization issues

korotkov-aerospike:


korotkov-aerospike:
1) use cache for s3 clients (will do other storage later)
2) use safe map to store jobs

pkg/service/backup_routine_handler.go
```diff
@@ -174,12 +180,15 @@ func (h *BackupRoutineHandler) runFullBackupInternal(ctx context.Context, now ti
...
-	h.clusterConfigWriter.Write(ctx, client.AerospikeClient(), now)
-
-	err = h.retentionManager.deleteOldBackups(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to clean up old backups: %w", err)
-	}
+	go func() {
+		// Cleanup old backups asynchronously.
+		// At this moment backup is already completed, but backupJob.isRunning flag is still set,
+		// we don't want to block other backup execution.
```
reugn:
```suggestion
		// Clean up old backups asynchronously.
		// At this moment backup is already completed, but backupJob.isRunning flag is still set,
		// potentially blocking subsequent backup executions.
```

pkg/service/backup_backend.go
```diff
@@ -178,6 +179,7 @@ func (b *BackupBackend) ReadClusterConfiguration(path string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	slog.Info("Reading cluster configuration", "routine", b.routineName, "path", path, "count", len(configBackups))
```
reugn: Consider using key-value pairs for logging.

pkg/util/safe_map.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// Delete removes a key from the map.
+func (s *SafeMap[K, V]) Delete(key K) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	delete(s.m, key)
+}
+
+func (s *SafeMap[K, V]) Apply(key K, callback func(value V)) {
```
reugn, line 39: Add documentation to methods where it is missing.

internal/server/handlers/backup.go
```diff
@@ -148,20 +148,18 @@ func (s *Service) readBackupsForRoutine(w http.ResponseWriter, r *http.Request,
 		return
 	}
 
+	hLogger = hLogger.With(slog.String("routine", routine))
```
reugn, line 151: Consider instrumenting only if required.

pkg/service/backup_scheduler.go
```diff
@@ -67,39 +59,37 @@ func NewScheduler(ctx context.Context) quartz.Scheduler {
...
-	clear(jobStore.jobs)
-
+	newJobs := map[string]*quartz.JobDetail{}
 	var errs error
 	for routineName, routine := range routines {
 		if routine.Disabled {
 			continue
 		}
-		handler := handlers[routineName]
+		handler, _ := handlers.Load(routineName)
```
reugn: Consider logging and continuing if not found, even if this scenario is not possible at this time.

internal/server/handlers/backup.go
korotkov-aerospike: yes, I added routine because it's used everywhere

reugn (approved):

PR #303, opened 2025-01-10
Fix race condition in client manager

korotkov-aerospike:


korotkov-aerospike:
The problem was that only checking of client existence was under lock. If it's absent, 2 threads both created client.

PR #308, opened 2025-01-16
APPS-1369 Optimize restore to empty databases

korotkov-aerospike:


korotkov-aerospike:
I've added finish time to restore job status to be able measure speed.
Testing of emptiness is incorporated into namespace validator, I don't want to create new struct.

pkg/dto/current_backup.go
```diff
@@ -44,8 +44,10 @@ type RunningJob struct {
 	TotalRecords uint64 `json:"total-records,omitempty" example:"100"`
 	// DoneRecords: the number of records that have been successfully done.
 	DoneRecords uint64 `json:"done-records,omitempty" example:"50"`
-	// StartTime: the time when the backup operation started.
+	// StartTime: the time when the operation started.
 	StartTime time.Time `json:"start-time,omitempty" example:"2006-01-02T15:04:05Z07:00"`
+	// FinishTime: the time when the operation finished
+	FinishTime *time.Time `yaml:"finish-time,omitempty" json:"finish-time,omitempty"`
```
reugn:
```suggestion
	FinishTime *time.Time `json:"finish-time,omitempty" example:"2006-01-02T15:04:05Z07:00"`
```

pkg/dto/restore_result.go
```diff
@@ -32,6 +24,10 @@ type RestoreStats struct {
 	FresherRecords  uint64 `yaml:"fresher-records,omitempty" json:"fresher-records,omitempty" format:"int64" example:"5"`
 	IndexCount      uint64 `yaml:"index-count,omitempty" json:"index-count,omitempty" format:"int64" example:"3"`
 	UDFCount        uint64 `yaml:"udf-count,omitempty" json:"udf-count,omitempty" format:"int64" example:"1"`
+
+	CurrentRestore *RunningJob `yaml:"current-restore,omitempty" json:"current-job,omitempty"`
+	Status         JobStatus   `yaml:"status,omitempty" json:"status,omitempty"`
+	Error          string      `yaml:"error,omitempty" json:"error,omitempty"`
```
reugn, line 30: What was the reason to unwrap?

pkg/service/aerospike/namespace_empty.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return false, fmt.Errorf("failed to get object count for set %s: %w", set, err)
+	}
+
+	return objectCount == 0, nil
+}
+
+// getNodeStats retrieves statistics from a node for a given key.
+func getNodeStats(node *as.Node, statsKey string) (string, error) {
+	infoRes, err := node.RequestInfo(&as.InfoPolicy{}, statsKey)
```
reugn: Could use `as.DefaultInfoPolicy`.

pkg/dto/restore_result.go
korotkov-aerospike: I just think it's unnecessary complication. We never use that nested struct independently, even in tests.

reugn (approved):

PR #320, opened 2025-02-13
Update github.com/aerospike/aerospike-client-go to v8 

korotkov-aerospike:
1) update version
2) wrappers (v8tov7 util class) for aerospike-management-lib
3) removed dependency on Aerospike/backup-go in some packages (moved under service/aerospike)
4) unified naming. Aerospike package is always "as", except files we use both versions (v7/v8)


go.mod
```diff
@@ -1,14 +1,15 @@
 module github.com/aerospike/aerospike-backup-service/v3
 
-go 1.22
+go 1.23.3
```
reugn:
```suggestion
go 1.23
```

reugn (dismissed):

reugn (approved):

PR #321, opened 2025-02-19
FMWK-675 Change config timeouts type to duration

korotkov-aerospike:


pkg/dto/backup_policy.go
```diff
@@ -128,19 +128,19 @@ func (p *BackupPolicy) ToModel() *model.BackupPolicy {
...
 }
 
-func durationToMillis(d *time.Duration) *int {
-	if d == nil {
+func durationToMillis(duration *time.Duration) *int64 {
+	if duration == nil {
 		return nil
 	}
-	ms := int((*d) / time.Millisecond)
+	ms := int64(*duration / time.Millisecond)
```
reugn: Could use `duration.Milliseconds()`.

reugn (approved):

PR #322, opened 2025-02-20
APPS-1503 Add support for specifying object storage classes

korotkov-aerospike:


pkg/dto/storage_gcp.go
```diff
@@ -58,5 +67,63 @@ func newGcpStorageFromModel(s *model.GcpStorage, config *model.BackupConfig) *Gc
...
+	GcpClassStandard,
+	GcpClassNearline,
+	GcpClassColdline,
+	GcpClassArchive,
+}
+
+// GcpStorageClass represents the configuration for GCP Storage Class.
+type GcpStorageClass struct {
+	// DataClass specifies the storage class for object data
+	DataClass string `json:"data" yaml:"data"`
```
reugn, line 95: Is there a `MetadataClass` for GCP Storage?

pkg/dto/storage_gcp.go
```diff
@@ -21,6 +23,8 @@ type GcpStorage struct {
 	// Alternative url.
 	// It is not recommended to use an alternate URL in a production environment.
 	Endpoint string `yaml:"endpoint,omitempty" json:"endpoint,omitempty"`
+	// StorageClass defines the storage class for data and metadata objects
```
reugn: Add a period to be consistent.

docs/dist/index.css
reugn, line 1: Why removing the Swagger UI files?

pkg/model/storage_local.go
```diff
@@ -6,13 +6,23 @@ import "fmt"
 // This interface is implemented by all specific storage types.
 type Storage interface {
 	GetPath() string
+	GetStorageClass() StorageClass
+}
+
+type StorageClass struct {
```
reugn, line 13: Add documentation.

pkg/dto/storage_gcp.go
korotkov-aerospike: https://cloud.google.com/storage/docs/storage-classes#classes

There is only one storage class (STANDARD) that is suitable for metadata (doest require retrival)

korotkov-aerospike:
@reugn  I tried GCP test in isolation and it also worked.
So, the problem with Automation tests is not related to Storage Class and should be fixed in other PR (probably in automation project only)

reugn (approved):

PR #336, opened 2025-03-31
APPS-1549 Improve restore path validation

korotkov-aerospike:
1) Added mockgen generator
2) new filter for backup backed service: PathFilter to search for backups in given path only
3) use this filter to validate backups before restore in data_restorer.


build/scripts/generate_mocks.sh
```diff
@@ -0,0 +1,57 @@
...
+if ! command -v mockgen > /dev/null 2>&1; then
+    go install go.uber.org/mock/mockgen@latest
+fi
+
+WORKSPACE="$(git rev-parse --show-toplevel)"
+echo "Workspace root: $WORKSPACE"
+
+# Define the base Go import path for your project
+# (Adjust if your module path is different)
+BASE_IMPORT_PATH="github.com/aerospike/aerospike-backup-service/v3"
```
reugn:
```suggestion
# Extract module path
BASE_IMPORT_PATH=$(grep '^module' "$WORKSPACE/go.mod" | awk '{print $2}')
echo "Module path: $BASE_IMPORT_PATH"
```

internal/server/handlers/restore.go
```diff
@@ -58,15 +37,21 @@ func (s *Service) RestoreFullHandler(w http.ResponseWriter, r *http.Request) {
...
+
+// RestoreIncremental and RestoreFull share same business logic.
+func (s *Service) restoreByPath(w http.ResponseWriter, r *http.Request) {
 	request, err := dto.NewRestoreRequestFromReader(r.Body)
 	if err != nil {
-		httpError(w, errBadRequest(err))
+		httpError(w, errInvalidJSONPayload(err))
 		return
 	}
 	if err = request.Validate(); err != nil {
```
reugn, line 50: Restore policy validation shouldn't allow nil value.

pkg/service/backup_backend_service.go
```diff
@@ -14,69 +14,129 @@ import (
...
 
-// WithToTime adds an end time to the filter.
-func (f BackupFilter) WithToTime(toTime time.Time) BackupFilter {
+func (f *RoutineFilter) WithToTime(toTime time.Time) *RoutineFilter {
 	f.ToTime = &toTime
 	return f
 }
 
-func (f BackupFilter) String() string {
+func (f RoutineFilter) getUpperBoundary() string {
```
reugn: Use pointer receiver to avoid mixing.

pkg/service/backup_backend_service.go
```diff
@@ -14,69 +14,129 @@ import (
...
 )
 
-// BackupFilter defines criteria for filtering backups.
-type BackupFilter struct {
-	// required
-	routine string
-	JobType jobType
+// BackupFilter is an interface that all filter types must implement.
+type BackupFilter interface {
+	isBackupFilter()
```
reugn: This construction is not idiomatic, but consider using something like
```go
filterType() filterType
```
instead, so that it can be used to identify the type instead of type matching.

pkg/service/backup_backend_service.go
```diff
@@ -114,33 +174,46 @@ func NewBackupBackendService(config *model.Config) *BackupBackendServiceImpl {
...
+	case *PathFilter:
+		return b.getPathBackups(ctx, *f)
+	default:
+		return nil, fmt.Errorf("unsupported filter type: %T", f)
+	}
+}
 
+func (b *BackupBackendServiceImpl) getRoutineBackups(
+	ctx context.Context,
+	filter RoutineFilter,
```
reugn: It is better to pass the filter as a pointer.

pkg/service/backup_backend_service.go
```diff
@@ -241,3 +316,34 @@ func (b *BackupBackendServiceImpl) Delete(ctx context.Context, routineName strin
 
 	return storage.DeleteFolder(ctx, routine.Storage, path)
 }
+
+func (b *BackupBackendServiceImpl) getPathBackups(
+	ctx context.Context,
+	filter PathFilter,
```
reugn: It is better to pass the filter as a pointer.

build/scripts/generate_mocks.sh
reugn, line 1: Other scripts use hyphens in their names.

internal/server/handlers/restore.go
korotkov-aerospike: why? We have default values for every field.

korotkov-aerospike:
```
func (p *RestorePolicy) ToModel() *model.RestorePolicy {
	if p == nil {
		return &model.RestorePolicy{}
	}

	return &model.RestorePolicy{
....
```

reugn (approved):

PR #337, opened 2025-04-02
APPS-1597 Allow 0 restore timeout

korotkov-aerospike:
By default it's still 10 minutes, but if user really want, he can set 0 (infinite)

### DIFF
--- a/pkg/dto/restore_policy.go
+++ b/pkg/dto/restore_policy.go
@@ -77,11 +77,11 @@ func (p *RestorePolicy) Validate() error {
 	if p.Parallel != nil && *p.Parallel <= 0 {
 		return errValidationNonPositive("parallel", *p.Parallel)
 	}
-	if p.TotalTimeout != nil && *p.TotalTimeout <= 0 {
-		return errValidationNonPositive("total-timeout", *p.TotalTimeout)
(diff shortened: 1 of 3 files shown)


pkg/dto/restore_policy.go
```diff
@@ -77,11 +77,11 @@ func (p *RestorePolicy) Validate() error {
...
 		return errValidationNonPositive("parallel", *p.Parallel)
 	}
-	if p.TotalTimeout != nil && *p.TotalTimeout <= 0 {
-		return errValidationNonPositive("total-timeout", *p.TotalTimeout)
+	if p.TotalTimeout != nil && *p.TotalTimeout < 0 {
+		return errValidationNegative("total-timeout", *p.TotalTimeout)
 	}
-	if p.SocketTimeout != nil && *p.SocketTimeout <= 0 {
-		return errValidationNonPositive("socket-timeout", *p.SocketTimeout)
+	if p.SocketTimeout != nil && *p.SocketTimeout < 0 {
```
reugn, line 83: Do we have a unit test for those?

korotkov-aerospike: no, you think we need? it's just simple validation

reugn (approved):

PR #345, opened 2025-04-20
APPS-1622 Fix several conversion bugs

korotkov-aerospike:


pkg/dto/compression_policy.go
```diff
@@ -29,9 +30,6 @@ func (p *CompressionPolicy) Validate() error {
 	if p.Mode != CompressNone && p.Mode != CompressZSTD {
 		return fmt.Errorf("invalid compression mode: %s", p.Mode)
 	}
-	if p.Level == 0 {
-		p.Level = -1
-	}
 	if p.Level < -1 {
```
reugn: It is worth validating the upper bound - 22.

reugn (approved):

PR #348, opened 2025-04-29
FMWK-742 Add configurable chunk size option to s3 writer

korotkov-aerospike:


pkg/service/storage/s3.go
```diff
@@ -52,6 +52,9 @@ func (a *S3StorageAccessor) createWriter(
 	if err != nil {
 		return nil, err
 	}
+	if s3s.MinPartSize > 0 {
+		opts = append(opts, ioStorage.WithChunkSize(s3s.MinPartSize))
```
reugn, line 56: Needs implementation for other storage providers as well.

korotkov-aerospike: I'll do it in separate tickets.

reugn (dismissed):

reugn (approved):

PR #350, opened 2025-05-07
APPS-1667 Add config JSON schema to allow autocomplete and real-time validation

korotkov-aerospike:


internal/server/configuration/configuration_manager.go
```diff
@@ -24,6 +25,9 @@ type Manager interface {
 	Write(ctx context.Context, config *model.Config) error
 }
 
+//nolint:lll
```
korotkov-aerospike, line 28: this will only start work after release

reugn (approved):

PR #354, opened 2025-05-11
Disable MRT backup functionality

korotkov-aerospike:


reugn (commented):
Need to remove the XDR section from `config.yml`.

reugn (approved):

PR #370, opened 2025-05-21
Automate metrics section generation in README

korotkov-aerospike:


korotkov-aerospike:
We forgot couple of metric in readme => they are not on website => users asks questions.

reugn (approved):

PR #373, opened 2025-05-25
APPS-1712 Add endpoint to query restore jobs details

korotkov-aerospike:


pkg/dto/restore_status_filter.go
```diff
@@ -0,0 +1,41 @@
...
+		s = strings.TrimSpace(s)
+		if s == "" {
+			continue
+		}
+
+		isExclude := strings.HasPrefix(s, "!")
+		if isExclude {
+			s = s[1:]
+		}
+
```
reugn:
```go
s = cases.Title(language.English).String(strings.ToLower(s))
```
To remove case-sensitivity in input.

pkg/model/restore_status_filter.go
```diff
@@ -0,0 +1,24 @@
+package model
+
+type StatusFilter struct {
```
reugn, line 4: Add a doc line explaining the structure purpose.

pkg/model/restore_status_filter.go
```diff
@@ -0,0 +1,24 @@
...
+		include: include,
+		exclude: exclude,
+	}
+}
+
+func (sf StatusFilter) Matches(status JobStatus) bool {
+	if sf.exclude[status] {
+		return false
+	}
+	if len(sf.include) > 0 {
```
reugn: Why checking the `len` first and only for the `include` map?

pkg/dto/restore_result.go
```diff
@@ -11,6 +15,24 @@ const (
...
 )
 
+var allJobStatuses = []JobStatus{
+	JobStatusRunning,
+	JobStatusDone,
+	JobStatusFailed,
+	JobStatusCancelled,
+}
+
+func RestoreStatusFromString(s string) (JobStatus, bool) {
```
reugn: * Can be private.
* Add a doc line.

pkg/dto/restore_result.go
```diff
@@ -11,6 +15,24 @@ const (
 	JobStatusCancelled JobStatus = "Cancelled"
 )
 
+var allJobStatuses = []JobStatus{
```
reugn, line 18: nit: could lowercase the job statuses once when instantiating the slice.

internal/server/handlers/restore.go
```diff
@@ -147,6 +147,42 @@ func extractJobID(r *http.Request) (model.RestoreJobID, error) {
...
+// @Param    from query int false "Lower bound timestamp filter" format(int64)
+// @Param    to query int false "Upper bound timestamp filter" format(int64)
+// @Param    status query string false "Comma-separated status filter (Running,Done,Failed,Cancelled). Use ! prefix for exclude filter (e.g., !Failed,Cancelled)"
+// @Router   /v1/restore/jobs [get]
+// @Success  200 {object} map[string]dto.RestoreJobStatus "Running restore jobs"
+// @Failure  400 {string} string
+// @Failure  500 {string} string
+//
+//nolint:lll
+func (s *Service) GetAllRestoreJobs(w http.ResponseWriter, r *http.Request) {
```
reugn: Renaming it to `RetrieveRestoreJobs` would be more accurate. It is not exactly `all` considering the filters.

reugn (approved):

PR #376, opened 2025-06-04
APPS-1692 Improve backup metrics

korotkov-aerospike:


README.md
```diff
@@ -270,18 +270,65 @@ following application metrics:
...
+| `aerospike_backup_service_incremental_failure_total`        | Counter   | Incremental backup failure counter (Deprecated, use `aerospike_backup_service_backup_events_total`)              |                        |
+| `aerospike_backup_service_duration_millis`                  | Counter   | Full backup duration in milliseconds (Deprecated, use `aerospike_backup_service_backup_duration_seconds`)        |                        |
+| `aerospike_backup_service_incremental_duration_millis`      | Counter   | Incremental backup duration in milliseconds (Deprecated, use `aerospike_backup_service_backup_duration_seconds`) |                        |
+| `aerospike_backup_service_backup_progress_pct`              | Gauge     | Progress of backup processes in percentage                                                                       | routine, type          |
+| `aerospike_backup_service_restore_progress_pct`             | Gauge     | Progress of restore processes in percentage                                                                      | job_id                 |
```
sunilvirus: We should revisit this aerospike_backup_service_restore_progress_pct. Especially, with the jobid that will be random. Prometheus does not like metrics which has too many unique labels (high-cardinality). We should consider this removing this altogether and allow only the rest api to query the status of an ongoing restore.

README.md
```diff
@@ -270,18 +270,65 @@ following application metrics:
...
+| `aerospike_backup_service_backup_events_total`              | Counter   | Backup service job events by routine, type, and outcome                                                          | routine, type, outcome |
+| `aerospike_backup_service_backup_duration_seconds`          | Histogram | Duration of backups by routine and type (in seconds)                                                             | routine, type          |
```
sunilvirus: This is a good one.

README.md
```diff
@@ -270,18 +270,65 @@ following application metrics:
...
+| `aerospike_backup_service_failure_total`                    | Counter   | Full backup failure counter (Deprecated, use `aerospike_backup_service_backup_events_total`)                     |                        |
+| `aerospike_backup_service_incremental_failure_total`        | Counter   | Incremental backup failure counter (Deprecated, use `aerospike_backup_service_backup_events_total`)              |                        |
+| `aerospike_backup_service_duration_millis`                  | Counter   | Full backup duration in milliseconds (Deprecated, use `aerospike_backup_service_backup_duration_seconds`)        |                        |
+| `aerospike_backup_service_incremental_duration_millis`      | Counter   | Incremental backup duration in milliseconds (Deprecated, use `aerospike_backup_service_backup_duration_seconds`) |                        |
+| `aerospike_backup_service_backup_progress_pct`              | Gauge     | Progress of backup processes in percentage                                                                       | routine, type          |
```
sunilvirus, line 295: In the docs, we should tell what are the possible values of type & outcome will be. i.e (full, incremental) & (success, fail, retry, skip). Give some examples also with outcome in your examples section.

README.md
```diff
@@ -270,18 +270,65 @@ following application metrics:
...
+| `aerospike_backup_service_restore_progress_pct`             | Gauge     | Progress of restore processes in percentage                                                                      | job_id                 |
+| `aerospike_backup_service_backup_events_total`              | Counter   | Backup service job events by routine, type, and outcome                                                          | routine, type, outcome |
+| `aerospike_backup_service_backup_duration_seconds`          | Histogram | Duration of backups by routine and type (in seconds)                                                             | routine, type          |
+| `aerospike_backup_service_last_successful_backup_timestamp` | Gauge     | Unix timestamp of the last successful backup per routine                                                         | routine                |
```
sunilvirus, line 299: Few more metrics that come to my mind as a user. Please discuss with @reugn too and make a final decision. 
1. How many recs were backed up in the ongoing/last run (labels : routine, type)
2.How many bytes were backed up in the ongoing/last run (labels : routine, type) 
3. Duration of ongoing/last backup. (labels : routine, type) 
4. How many backups are retained (based on our retention policy) (labels : routine, type)
5. How many backups routines are running currently. Can I derive this from aerospike_backup_service_backup_progress_pct ? sum of all routines with 1 <= pct < 100 ? If a backup fails after 50%, do we reset this counter to 0 ? 
6. How many total scans on aerospike are ongoing from ABS (will be a nice thing to alert on) (label: Aerospike-cluster)

On the cloud storage side...
1. Number of retries to the cloud storage (label: storage). This happens occasionally. Right ?
2. Number of write flushes to the cloud storage (after our record grouping)
3. Histogram of cloud write (in ms). I think this will be very useful especially to troubleshoot issues when cloud storage is the bottleneck. 
4. Any other cloud storage metrics you can think of ?

korotkov-aerospike:
@sunilvirus @reugn I had impletemented all review comments, except the last one (the biggest)
will do it in separate ticket.

reugn (approved):

pkg/service/prometheus.go
```diff
@@ -67,12 +67,45 @@ var (
 		[]string{"routine", "type"},
 	)
 	// A gauge metrics for restore processes.
-	restoreProgress = prometheus.NewGaugeVec(
+	restoreInProgress = prometheus.NewGauge(
 		prometheus.GaugeOpts{
 			Name: "aerospike_backup_service_restore_progress_pct",
```
sunilvirus, line 72: @korotkov-aerospike I see that you changed the behaviour to not have per restore-job info but only count of ongoing restore jobs. That is good. But the name is still aerospike_backup_service_restore_progress_pct. The name is misleading as it says pct (percentage). We should rename the metrics also to something like aerospike_backup_service_restore_in_progress or aerospike_backup_service_restore_active

korotkov-aerospike: right, sorry, I renamed variable but forgot the name. Will fix it.

PR #382, opened 2025-06-09
APPS-1654 Add more robust storage retry strategy

korotkov-aerospike:


pkg/model/config_default.go
```diff
@@ -70,3 +70,15 @@ var defaultConfig = struct {
 		},
 	},
 }
+
+var StorageRetryPolicy = struct {
```
reugn, line 76: Document the purpose of this policy variable.

reugn (approved):

PR #388, opened 2025-06-22
APPS-1766 Fix counter for incremental backup

korotkov-aerospike:


korotkov-aerospike:
Also, I added 2 small fixes from other PRs.

pkg/service/backup_routine_orchestrator.go
```diff
@@ -208,7 +208,7 @@ func (h *BackupRoutineOrchestrator) runIncrementalBackup(ctx context.Context, no
 		observeBackupEvent(h.routineName, jobTypeIncremental, BackupOutcomeFailure, duration)
 		h.logger.Error("Incremental backup failed", slog.Any("error", err))
 	} else {
-		observeBackupEvent(h.routineName, jobTypeIncremental, BackupOutcomeFailure, duration)
+		observeBackupEvent(h.routineName, jobTypeIncremental, BackupOutcomeSuccess, duration)
```
reugn, line 211: Is there any test to validate the metric?

pkg/service/prometheus.go
```diff
@@ -66,10 +66,10 @@ var (
 		},
 		[]string{"routine", "type"},
 	)
-	// A gauge metrics for restore processes.
+	// A gauge metric for restore processes.
```
reugn: Same thing for the previous metric.

korotkov-aerospike: the previous is an array, so plural

reugn (approved):

PR #391, opened 2025-06-24
APPS-1780 Fix validation for records per second

korotkov-aerospike:


reugn (approved):

korotkov-aerospike:
I want to update the backup lib (low bandwidth fix) before merge

PR #392, opened 2025-06-26
Use Alpine as base Docker image; update GH workflows

davi17g:


reugn (dismissed):

.snyk
reugn, line 1: The file is empty.

davi17g: yes, you specify in this file some known vulnerabilities that you want snyk to skip.

reugn (approved):

PR #393, opened 2025-06-29
APPS-1779 Enforce and validate minimum bandwidth

korotkov-aerospike:


pkg/service/restore_data.go
```diff
@@ -102,6 +103,8 @@ func (r *dataRestorer) executeRestore(
 	r.restoreJobs.addHandler(jobID, handler)
 
 	// Wait for the restore operation to complete
+	slog.Info("wait", slog.Any("jobID", jobID))
```
reugn: > wait for restore job

pkg/dto/constants.go
```diff
@@ -3,4 +3,8 @@ package dto
 const (
 	// max possible value https://aerospike.com/docs/server/reference/configuration#namespace__rack-id
 	maxRack = 1000000
+
+	// We need to enforce a minimum bandwidth due to rate limiter constraints.
+	// Set to 8 MiB/s for the maximum record size.
+	minBandwidth = 8
```
reugn, line 9: It states that this is the minimal value for bandwidth, but we require at least 9 here.

reugn (approved):

PR #400, opened 2025-07-06
APPS-1796 Skip empty backups on restore

korotkov-aerospike:


pkg/service/restore_data.go
```diff
@@ -93,7 +93,14 @@ func (r *dataRestorer) executeRestore(
 		return fmt.Errorf("failed to read backups: %w", err)
 	}
 
-	if err := r.validateBackupData(backups); err != nil {
+	if len(backups) > 0 && r.allBackupsEmpty(backups) {
+		// edge case: backups exist but are empty — nothing to restore.
```
reugn, line 97: Worth adding an info log for the condition.

reugn (approved):

PR #403, opened 2025-07-13
Set minimal permissions in GitHub workflows

reugn:


.github/workflows/snyk-scan.yml
```diff
@@ -17,11 +21,12 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run Snyk to check for vulnerabilities
-        uses: snyk/actions/golang@master
+        uses: snyk/actions/golang@cdb760004ba9ea4d525f2e043745dfe85bb9077e
         continue-on-error: true # To make sure that SARIF upload gets called
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
+          snyk-version: v1.1297.3
```
davi17g, line 29:
```suggestion
          snyk-version: latest
```
as far as I know it is recommended always to have the latest cli

.github/workflows/snyk-scan.yml
```diff
@@ -17,11 +21,12 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run Snyk to check for vulnerabilities
-        uses: snyk/actions/golang@master
+        uses: snyk/actions/golang@cdb760004ba9ea4d525f2e043745dfe85bb9077e
```
davi17g, line 24: Would be great if there cloud be the comment with version tag of that commit as it is done here for example:
`uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2`

davi17g (approved):

PR #407, opened 2025-07-21
Use exact trigger timestamps in backup jobs

korotkov-aerospike:


pkg/util/cron.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"time"
+
+	"github.com/reugn/go-quartz/quartz"
+)
+
+// IsCronFireTime returns true if the given timestamp exactly matches a scheduled fire time.
+func IsCronFireTime(cronExpr string, t time.Time) bool {
+	trigger, _ := quartz.NewCronTrigger(cronExpr)
```
reugn: Consider checking the error and returning false if it is not nil.

pkg/service/backup_routine_orchestrator.go
```diff
@@ -214,24 +214,32 @@ func (h *BackupRoutineOrchestrator) runIncrementalBackup(ctx context.Context, no
...
+			h.logger.Debug("Skipping incremental backup: another incremental backup in progress")
+			return true
+		}
+		if util.IsCronFireTime(h.routine.IntervalCron, now) {
+			h.logger.Debug("Skipping incremental backup: full backup scheduled at same time")
+			return true
+		}
 	}
 
 	return false
```
reugn: Can be simplified to:
```go
        // Concurrent incremental are only allowed when explicitly set (default is false).
	if allowConcurrent := h.routine.BackupPolicy.ConcurrentIncremental != nil &&
		*h.routine.BackupPolicy.ConcurrentIncremental; allowConcurrent {
		return false
	}

	switch {
	case currentStat.Full != nil:
		h.logger.Debug("Skipping incremental backup: full backup in progress")
	case currentStat.Incremental != nil:
		h.logger.Debug("Skipping incremental backup: another incremental backup in progress")
	case util.IsCronFireTime(h.routine.IntervalCron, now):
		h.logger.Debug("Skipping incremental backup: full backup scheduled at same time")
	}

	return true
```

pkg/service/backup_job.go
```diff
@@ -33,13 +32,16 @@ var _ quartz.Job = (*backupJob)(nil)
 
 // Execute is called by a Scheduler when the Trigger associated with this job fires.
 func (j *backupJob) Execute(ctx context.Context) error {
+	jobMetadata, _ := ctx.Value(quartz.JobMetadataContextKey).(quartz.JobMetadata)
```
reugn: It is better to check whether the type assertion succeeded and return an error otherwise.

reugn (approved):

PR #415, opened 2025-07-28
APPS-1805 Improve TLS validation

korotkov-aerospike:


pkg/dto/tls.go
```diff
@@ -0,0 +1,147 @@
...
+		return errValidationRequires("key-file-password", "key-file")
+	}
+
+	return nil
+}
+
+// validateTLSConfig attempts to create a TLS config to catch low-level issues.
+func (t *TLS) validateTLSConfig() error {
+	if _, err := aerospike.NewTLSConfig(t.toModel()); err != nil {
+		return fmt.Errorf("tls: invalid configuration: %w", err)
```
reugn:
```suggestion
		return fmt.Errorf("tls %w: %w", errValidation, err)
```

pkg/dto/aerospike_cluster.go
```diff
@@ -59,6 +64,26 @@ func (a *AerospikeCluster) Validate() error {
 	return nil
 }
 
+// validateSeedNodesTLSConsistency ensures that if any seed node has TLS configuration,
+// then all seed nodes must have TLS configuration.
+func (a *AerospikeCluster) validateSeedNodesTLSConsistency() error {
+	var hasTLSNodes, hasNonTLSNodes bool
```
reugn, line 74:
```suggestion
	if len(a.SeedNodes) <= 1 {
		return nil
	}
	
	var hasTLSNodes, hasNonTLSNodes bool
```
To optimize for a single seed node configuration.

reugn (approved):

PR #416, opened 2025-07-29
Remove unused StorageConfig from RestoreTimestampRequest DTO

korotkov-aerospike:


korotkov-aerospike:
It's just in DTO, actually storage is taken from the backup

reugn (approved):

PR #422, opened 2025-08-11
APPS-1856 Use backup library client for info requests

korotkov-aerospike:


korotkov-aerospike:
0. Info client is taken from the backup client. Interfaces are used instead of backp.Client
1. moved ns validation out of toModel. My initial intent was to have always valid model, but the price of making network call during conversion is too high.
2. ns validation is not an error, just a log entry. Users might want to configure system offline. THe error is when we have no NS during backup/restore only.

pkg/model/config_default.go
```diff
@@ -84,3 +84,9 @@ var StorageRetryPolicy = struct {
 	},
 	MaxDuration: 2 * time.Minute,
 }
+
+var InfoDefaultRetryPolicy = &models.RetryPolicy{
```
reugn: Document this variable.

pkg/service/aerospike/client_manager.go
```diff
@@ -47,7 +41,7 @@ type ClientManagerImpl struct {
 var _ ClientManager = (*ClientManagerImpl)(nil)
 
 type clientInfo struct {
-	client     *backup.Client
+	client     Client
```
reugn, line 44: Consider holding `backup.InfoGetter` instead.

pkg/service/aerospike/namespace_validator.go
```diff
@@ -1,132 +1,120 @@
...
 }
 
-func (nv *defaultNamespaceValidator) MissingNamespaces(
-	cluster *model.AerospikeCluster,
-	namespaces []string,
-) []string {
-	if len(namespaces) == 0 {
-		return nil
-	}
+type MissingByRoutine map[string][]string
```
reugn: Document this type.

pkg/service/aerospike/client_manager.go
korotkov-aerospike: I need the full Client, because that's what I return in GetClient

pkg/model/config_default.go
korotkov-aerospike: removed it.

reugn (dismissed):

reugn (approved):

PR #423, opened 2025-08-12
APPS-1872 Print logs on routine schedule

korotkov-aerospike:


pkg/service/backup_scheduler.go
```diff
@@ -119,6 +117,12 @@ func schedule(scheduler Scheduler, interval string, jobDetail *quartz.JobDetail)
 		return err
 	}
 
+	fireTime, err := cronTrigger.NextFireTime(time.Now().UnixNano())
+	if err != nil {
+		return err
+	}
+	jobDetail.Job().(*backupJob).logger.Info("Schedule", slog.Any("nextRun", time.Unix(0, fireTime)))
```
reugn: * Use safe type assertion.
* Consider printing the next run time in a user friendly format.

pkg/service/backup_job.go
```diff
@@ -64,7 +64,7 @@ func (j *backupJob) Description() string {
 }
 
 // newBackupJob creates a new backup job.
-func newBackupJob(runner backupRunner, jobType jobType, routineName string) quartz.Job {
+func newBackupJob(runner backupRunner, jobType jobType, routineName string) *backupJob {
```
reugn: This change is unnecessary.

pkg/service/backup_scheduler.go
korotkov-aerospike: It's friendly by default

time=2025-08-12T15:03:29.320+03:00 level=INFO source=/Users/ak/work/aerospike-backup-service/pkg/service/backup_scheduler.go:124 msg=Schedule routine=incrementalBackupEncryptedCompressed128 type=full nextRun=2026-01-01T02:00:00.000+02:00

pkg/service/backup_job.go
korotkov-aerospike: yeah, I tried to avoid type asserion; will revert

reugn (approved):

PR #432, opened 2025-09-10
APPS-1914 Show estimated end time only for ongoing restores

korotkov-aerospike:


pkg/service/estimates.go
```diff
@@ -39,8 +39,11 @@ func RestoreJobStatus(job *restoreJob) *model.RestoreJobStatus {
 	restoreStats := models.SumRestoreStats(stats...)
 
 	doneRecords := restoreStats.GetReadRecords()
-	runningJob := NewRunningJob(job.started, job.finished, doneRecords, job.totalRecords)
-	runningJob.Metrics = models.SumMetrics(metrics...)
+	var runningJob *model.RunningJob
+	if job.status == model.JobStatusRunning { // Metrics are only relevant for running restores.
```
reugn, line 43: It is likely a data race here.

reugn: Agreed to be addressed in a separate PR along with other similar issues.

reugn (approved):

PR #436, opened 2025-09-28
APPS-1921 Expose scan policy configuration properties

korotkov-aerospike:


pkg/dto/backup_policy.go
```diff
@@ -64,7 +64,13 @@ type BackupPolicy struct {
 
 	// Allows incremental backups to run concurrently.
 	// When false (default), incremental backups are skipped if another backup for same routine is in progress.
-	ConcurrentIncremental *bool `yaml:"concurrent-incremental,omitempty" json:"concurrent-incremental,omitempty" default:"false"`
+	ConcurrentIncremental *bool `yaml:"concurrent-incremental,omitempty" json:"concurrent-incremental,omitempty" extensions:"x-nullable"`
+	// Enables built-in compression during scan operation.
+	// Valid for Aerospike Server Enterprise Edition only.
+	UseCompression *bool `yaml:"use-scan-compression,omitempty" json:"use-scan-compression,omitempty" extensions:"x-nullable"`
+	// Maximum number of concurrent requests to server nodes.
+	// Default (0) is to issue requests to all server nodes in parallel.
```
reugn:
```suggestion
	// Default is to issue requests to all server nodes in parallel.
```
Since the filed is a pointer and its default value is nil (requires regenerating the OpenAPI spec).

reugn (approved):

PR #438, opened 2025-10-15
APPS-1920 Add support for optional fields in retry policy

korotkov-aerospike:


pkg/dto/retry_policy.go
```diff
@@ -28,45 +25,46 @@ func (r *RetryPolicy) Validate() error {
 		return nil
 	}
 
-	if r.BaseTimeout <= 0 {
-		return errValidationNonPositive("base-timeout", r.BaseTimeout)
+	if r.BaseTimeout != nil {
+		if *r.BaseTimeout > maxTimeout {
+			return errValidationInvalidValue("base-timeout", *r.BaseTimeout, "should not exceed 24 hours")
```
reugn: "should not exceed 24 hours" can be extracted into a `maxTimeoutMessage` and reused.

pkg/model/retry_policy.go
```diff
@@ -0,0 +1,50 @@
...
+	// The actual delay is calculated as: BaseTimeout * (Multiplier ^ attemptNumber)
+	Multiplier *float64
+
+	// MaxRetries is the maximum number of retry attempts that will be made.
+	// If set to 0, no retries will be performed.
+	MaxRetries *int
+}
+
+// Restore Retry policy used for write operations on restore.
+func (policy *RetryPolicy) Restore() *models.RetryPolicy {
```
reugn, line 24: Consider checking the receiver for `nil` in the methods.

pkg/dto/retry_policy.go
```diff
@@ -1,25 +1,22 @@
...
-	BaseTimeout int64 `json:"base-timeout" yaml:"base-timeout"`
+	BaseTimeout *int64 `json:"base-timeout,omitempty" yaml:"base-timeout,omitempty" extensions:"x-nullable"`
 
 	// Multiplier is used to increase the delay between subsequent retry attempts.
 	// The actual delay is calculated as: BaseTimeout * (Multiplier ^ attemptNumber)
-	Multiplier float64 `json:"multiplier" yaml:"multiplier"`
+	Multiplier *float64 `json:"multiplier,omitempty" yaml:"multiplier,omitempty" extensions:"x-nullable"`
 
 	// MaxRetries is the maximum number of retry attempts that will be made.
 	// If set to 0, no retries will be performed.
```
reugn, line 18: The validation returns an error if set to 0.

pkg/model/retry_policy.go
korotkov-aerospike: I do check

	if policy != nil && policy.BaseTimeout != nil {
		timeout = *policy.BaseTimeout
	}

or it's not what you mean?

pkg/dto/retry_policy.go
korotkov-aerospike: if r.MaxRetries != nil && *r.MaxRetries < 0 {
		return errValidationNegative("max-retries", *r.MaxRetries)
	}

should pass for 0

pkg/model/retry_policy.go
reugn: Right, it is checked in `toLibraryModels`.

reugn (approved):

PR #439, opened 2025-10-16
APPS-1871 Cap percentage values at 100%

korotkov-aerospike:


pkg/service/estimates.go
```diff
@@ -64,7 +64,7 @@ func NewRunningJob(
 		}
 	}
 
-	percentage := float64(done) / float64(total)
+	percentage := min(float64(done)/float64(total), 1.0) // percentage should not exceed 100%.
```
reugn: As we discussed, the counter should stop at 99% and only update to 100% once the processing is completed.

reugn (approved):

PR #442, opened 2025-10-20
APPS-1885 Add human-readable date to backup file paths

korotkov-aerospike:


config/config.yml
```diff
@@ -276,3 +276,4 @@ service:
     level: INFO
     file-writer:
       filename: "log/backup-service.log"
+  date-format: EU
```
reugn: It is better to create another subsection for backup configuration, e.g.:
```yaml

service:
  logger:
    level: INFO
    file-writer:
      filename: "log/backup-service.log"
  backup:
    timestamp-format: EU
```
If not specified or (NONE?) is chosen, no formatting is applied, keeping the existing timestamp.

pkg/dto/backup_service_config.go
```diff
@@ -14,19 +17,50 @@ type BackupServiceConfig struct {
 	HTTPServer *HTTPServerConfig `yaml:"http,omitempty" json:"http,omitempty"`
 	// Logger is the backup service logger configuration.
 	Logger *LoggerConfig `yaml:"logger,omitempty" json:"logger,omitempty"`
+	// Encoding for backup date in human-readable format (optional)
+	// Allowed values:
+	// * ISO (e.g. 2006-01-02T15-04-05)
+	// * EU (e.g. 02-Jan-2006-15-04-05)
+	// * US (e.g. Jan-02-2006-15-04-05)
```
reugn: EU and US formats are not sortable. Make sure there are unit tests verifying their correctness.

pkg/model/backup_service_config.go
```diff
@@ -6,6 +6,8 @@ type BackupServiceConfig struct {
 	HTTPServer *HTTPServerConfig
 	// Logger is the backup service logger configuration.
 	Logger *LoggerConfig
+	// DateFormat for human-readable dates in path.
```
reugn:
```suggestion
	// DateFormat for human-readable dates in backup file paths.
```

reugn (approved):

PR #444, opened 2025-10-21
Remove explicit request body closures in handlers

reugn:
For server requests, the HTTP server handles this automatically. See the [doc](https://pkg.go.dev/net/http#Request).

### DIFF
--- a/internal/server/handlers/config_cluster.go
+++ b/internal/server/handlers/config_cluster.go
@@ -32,17 +32,15 @@ func (s *Service) AddAerospikeCluster(w http.ResponseWriter, r *http.Request) {
 		httpError(w, errInvalidJSONPayload(err))
 		return
 	}
-	r.Body.Close()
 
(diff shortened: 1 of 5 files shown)


internal/server/handlers/config_cluster.go
```diff
@@ -32,17 +32,15 @@ func (s *Service) AddAerospikeCluster(w http.ResponseWriter, r *http.Request) {
 		httpError(w, errInvalidJSONPayload(err))
 		return
 	}
-	r.Body.Close()
 
-	err = s.changeConfig(r.Context(), func(config *model.Config) error {
+	if err = s.changeConfig(r.Context(), func(config *model.Config) error {
```
korotkov-aerospike, line 36: I find lambda inside of an if statement is too much nesting, it's harder to read then current version

korotkov-aerospike (approved):

PR #447, opened 2025-10-28
Update `aerospike-client-go` and `aerospike-management-lib` versions

korotkov-aerospike:


go.mod
```diff
@@ -1,15 +1,15 @@
 module github.com/aerospike/aerospike-backup-service/v3
 
-go 1.23.0
```
korotkov-aerospike, line 3: cause of aerospike-management-lib v1.8.0

filkeith (approved):

PR #450, opened 2025-10-29
APPS-1943 Allow validation to skip TLS files

korotkov-aerospike:


pkg/dto/validate_options.go
```diff
@@ -0,0 +1,8 @@
+package dto
+
+// ValidationOption holds configuration for validation behavior.
+type ValidationOption int
+
+const (
+	SkipTLSFiles ValidationOption = iota
```
filkeith: Consider add some prefix to option, so in future the will be a bunch of options, and it will be easier to understand what this constant contain.

Smth like:

```go
const (
  OptionSkipTLSFiles ValidationOption = iota
  ...
  OptionSkipSmthElse
  OptionApplySmth
)
```

Or may be just Opt.... like `OptSkipTLSFiles`

filkeith (approved):
LGTM

PR #451, opened 2025-10-29
APPS-1945 Fix client-policy for rack-list backup

filkeith:
- if rack list is set we should use scanPolicy.ReplicaPolicy = as.MASTER

### DIFF
--- a/pkg/service/backupexecutor/scan.go
+++ b/pkg/service/backupexecutor/scan.go
@@ -112,7 +112,7 @@ func scanPolicy(
 		scanPolicy.ReplicaPolicy = as.PREFER_RACK
 	}
 
-	if len(backupRoutine.RackList) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/service/backupexecutor/scan.go
```diff
@@ -112,6 +112,10 @@ func scanPolicy(
 		scanPolicy.ReplicaPolicy = as.PREFER_RACK
 	}
 
+	if len(backupRoutine.RackList) > 0 || len(backupRoutine.NodeList) > 0 {
```
korotkov-aerospike, line 115: could you please add validation for it to unit test (and also for backupRoutine.SourceCluster.PreferRacks, I forgot)

filkeith: sure

korotkov-aerospike (approved):

PR #460, opened 2025-11-12
APPS-1916 Make service.restoreJob thread safe

korotkov-aerospike:


pkg/service/jobs_holder.go
```diff
@@ -31,20 +34,75 @@ type restoreJob struct {
...
+		j.status = model.JobStatusFailed
+	}
+}
+
+// buildStatus constructs a model.RestoreJobStatus from the current job state.
+func (j *restoreJob) buildStatus() *model.RestoreJobStatus {
+	j.RLock()
+	defer j.RUnlock()
+
+	return RestoreJobStatus(j)
```
reugn: Make `RestoreJobStatus` a private method of `restoreJob`.

pkg/service/jobs_holder.go
```diff
@@ -31,20 +34,75 @@ type restoreJob struct {
...
+	j.Lock()
+	defer j.Unlock()
+
+	j.finished = ptr.Of(time.Now())
+	j.err = err
+
+	switch {
+	case err == nil:
+		j.status = model.JobStatusDone
+	case errors.Is(err, context.Canceled):
```
reugn, line 91: Is there a possibility for `context.DeadlineExceeded`?

korotkov-aerospike: that will fall into Failed category.

reugn (approved):

PR #468, opened 2025-12-03
APPS-1836 Ensure immutable routine configuration in backup and restore

korotkov-aerospike:
This PR refactors the service layer to pass *model.BackupRoutine objects instead of routine names (string) across the application (Orchestrator, Registry, Backend Service, Retention Manager).

  Key Changes:
   * Contextual Data Flow: Services now operate on self-contained BackupRoutine objects, reducing the need for repeated configuration lookups and locking.
   * Model Update: model.BackupRoutine now includes its own Name field.
   * Concurrency Safety: Added a Copy() method to BackupRoutine to provide safe, immutable snapshots of routine configurations to async processes.
   * Registry Decoupling: RunningBackupsRegistry now consumes a narrower routineProvider interface instead of the entire Config object.
   * Fix: Added validation in PopInvalidatedRoutines to ensure deleted routines do not cause nil pointer exceptions.


korotkov-aerospike:
Tests passed https://automation.ecoeng.dev/jenkins/job/abs/job/test-abs-locally/job/gcp-aws-setup/12/console

reugn (approved):

PR #469, opened 2025-12-08
APPS-1836 Distinguish between nil and 0 on copy

korotkov-aerospike:


pkg/util/optional/optional.go
reugn, line 1: Consider taking the package outside of `collections` and renaming to something like `monad`.

reugn (approved):

PR #473, opened 2025-12-16
APPS-2006 Add storage connectivity checks

korotkov-aerospike:
Features
   * Storage Connectivity: Added connectivity checks for S3, Azure, and GCP storage backends. The service now verifies access to the configured bucket/container upon initialization, failing fast if the storage is unreachable.
   * Azure Resilience: Implemented the global storage retry policy (max retries, delay, backoff) for the Azure Blob Storage client (was missing previously for some auth types)
   * Azure Auth are using pointer receivers now, to pass gob serialization-deserialization.


pkg/service/storage/azure.go
```diff
@@ -132,10 +154,31 @@ func clientWithDefaultCredential(endpoint string) (*azblob.Client, error) {
 		return nil, fmt.Errorf("failed to obtain default Azure credentials: %w", err)
 	}
 
-	client, err := azblob.NewClient(endpoint, cred, nil)
+	client, err := azblob.NewClient(endpoint, cred, &azblob.ClientOptions{
+		ClientOptions: azcore.ClientOptions{
```
filkeith, line 151: What do you think about making 1 function with opts and call it all CLients. Smth like
```
func getOpts() azcore.ClientOptions {
 azcore.ClientOptions{
			 Retry: policy.RetryOptions{
			 	MaxRetries:    int32(model.StorageRetryPolicy.MaxRetries),
			 	RetryDelay:    model.StorageRetryPolicy.BaseTimeout,
				MaxRetryDelay: model.StorageRetryPolicy.MaxBackoffDuration,
			},
		}
}
```

So if smth will change you can update all auth at once

reugn (approved):

PR #484, opened 2026-02-04
APPS-2031 refactor components initialization 

korotkov-aerospike:
The PR does 2 things:

Introduces BackupCompletionHandler – Post-success/failure behavior (registry, retention, cluster config) is moved behind an interface in pkg/service/backup_completion.go. This simplifies BackupRoutineOrchestrator and BackupComponents – The orchestrator and components depend on BackupCompletionHandler instead of RetentionManager and ClusterConfigWriter; tests use MockBackupCompletionHandler.

Moves wiring to internal/app – initComponents is moved from cmd/backup/main.go into internal/app;


korotkov-aerospike:
regression passed https://automation.ecoeng.dev/jenkins/job/abs/job/test-abs-locally/job/master/4259/console

internal/app/bootstrap.go
```diff
@@ -0,0 +1,134 @@
...
+	u "github.com/aerospike/aerospike-backup-service/v3/pkg/util/collections"
+	"github.com/reugn/go-quartz/quartz"
+)
+
+// InitComponents builds the full object graph and returns config, scheduler, HTTP service, and logger.
+func InitComponents(
+	ctx context.Context,
+	configFile string,
+	remote bool,
+) (*model.Config, quartz.Scheduler, *handlers.Service, *slog.Logger, error) {
```
filkeith: Are you sure that you need a func that returns 5 objects?
It's better to split it in 3-4 functions or return struct with all thi feilds

filkeith (changes requested):

filkeith (approved):
LGTM

PR #491, opened 2026-02-15
BRKS-21 Use only finished backups on restore

korotkov-aerospike:


pkg/service/backend_reader.go
```diff
@@ -0,0 +1,150 @@
...
+// keyFromStoragePath returns the backup key (path to dir containing metadata.yaml, relative to storage).
+func keyFromStoragePath(storagePath string) string {
+	return filepath.Dir(storagePath)
+}
+
+// filterBackups returns backups whose Created and Finished times fall within bounds.
+func (r *backupReader) filterBackups(backups []model.BackupDetails, bounds model.TimeBounds) []model.BackupDetails {
+	out := make([]model.BackupDetails, 0, len(backups))
+	for _, b := range backups {
+		if bounds.Contains(b.Created) && bounds.Contains(b.Finished) {
```
filkeith, line 126: Won't zero value of b.Finished brake this check?

pkg/service/backend_reader.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	return out
+}
+
+// filterEligibleFiles returns file paths that are at or before the path upper bound.
+// File paths and maxPath are relative to storage root (same format ReadFile expects).
+func (r *backupReader) filterEligibleFiles(files []string, maxPath string) []string {
+	out := make([]string, 0, len(files))
+	for _, p := range files {
+		if p < maxPath || strings.HasPrefix(p, maxPath) {
```
filkeith, line 81: Do you sort files before this check p < maxPath? If files are not sorted you can loose something

internal/server/handlers/backup.go
```diff
@@ -165,10 +165,6 @@ func (s *Service) readAllBackups(
 	for _, routine := range s.config.Routines() {
 		routineBackups, err := s.readBackupsForRoutine(ctx, filter(routine))
 		if err != nil {
-			if errors.Is(err, service.ErrNotFound) {
-				continue // Not an error. Routine may have been deleted while reading backups for previous routines.
-			}
-
 			return nil, err
```
filkeith, line 168: Won't it brake backward compatibility, as we now return NotFound error

pkg/service/backend_filter.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// WithFromTime sets the lower bound for Created timestamp filtering.
+func (f *RoutineFilter) WithFromTime(fromTime time.Time) *RoutineFilter {
+	f.FromTime = &fromTime
+	return f
+}
+
+// WithFromTime sets the lower bound for Created timestamp filtering.
+func (p *PathFilter) WithFromTime(fromTime time.Time) *PathFilter {
```
filkeith: Move this to PathFilter

pkg/service/backend_reader.go
korotkov-aerospike: we dont have zero for b.Finished
not finished backups dont exist

pkg/service/backend_reader.go
korotkov-aerospike: they are not sorted, but also I dont stop on reaching maxPath, I filter all of them anyway

filkeith (dismissed):

internal/server/handlers/backup.go
korotkov-aerospike: no, it's leftover from change long ago.
We dont have service.ErrNotFound anymore.
this was from them times when I pass the routine name to the filter, and it might be missing.
But now I pass the routine itself.

pkg/service/backend_reader.go
korotkov-aerospike: @filkeith But it is a good point anyway, we should explicitly sort files

filkeith (approved):
lgtm

PR #503, opened 2026-02-17
BKRS-81 Add explicit cancelled outcome and logging for backup runs

korotkov-aerospike:
PR adds explicit cancelled outcome handling and related logging/docs updates.


filkeith (approved):
lgtm

PR #505, opened 2026-02-18
BRKS-12 Add ability to trigger incremental backup via API

korotkov-aerospike:


filkeith (dismissed):
lgtm

korotkov-aerospike:
I found a small nil dereference during automation tests; it's ok now

pkg/model/backup_policy.go
```diff
@@ -136,6 +136,14 @@ func (p *BackupPolicy) GetFileLimitOrDefault() int {
 	return *defaultConfig.backupPolicy.FileLimit
 }
 
+func (p *BackupPolicy) GetSocketTimeoutOrDefault() time.Duration {
+	if p != nil && p.SocketTimeout != nil && *p.SocketTimeout > 0 {
```
filkeith: But if p.SocketTimeout = 0 it will be ignored, and DefaultSocketTimeout will be returned

filkeith (changes requested):

pkg/model/backup_policy.go
korotkov-aerospike: right

filkeith (approved):

PR #506, opened 2026-02-18
Reduce Dependabot noise with monthly grouped updates

korotkov-aerospike:
This PR reduces dependency-update noise and applies the latest Go dependency updates in one consolidated change, while fixing a code-level deprecation  surfaced by those upgrades.

Fixed staticcheck deprecations in pkg/service/storage/gcp.go caused by updated deps


filkeith (approved):
Need to wait month to test iof everything works

PR #507, opened 2026-02-19
BRKS-86 Refactor data restorer component

korotkov-aerospike:
This PR does almost not change the functionality

 Split logic into two independent runner files:
    • pkg/service/restore_path_runner.go
    • pkg/service/restore_time_runner.go

  • Updated NewRestoreManager(...) to construct and inject both runners directly.
  • Changed behavior:
    • restore-by-time also validates destination namespace
    • removed duplicate finishJob call in path empty-backup edge case

### DIFF
--- a/pkg/service/restore_manager.go
+++ b/pkg/service/restore_manager.go
@@ -2,8 +2,16 @@ package service
 
 import (
 	"context"
+	"fmt"
+	"log/slog"
+	"slices"
 
 	"github.com/aerospike/aerospike-backup-service/v3/pkg/model"
+	"github.com/aerospike/aerospike-backup-service/v3/pkg/service/aerospike"
+	"github.com/aerospike/aerospike-backup-service/v3/pkg/service/restoreexecutor"
+	"github.com/aerospike/aerospike-backup-service/v3/pkg/util/collections"
+	"github.com/aerospike/backup-go"
+	"github.com/aerospike/backup-go/models"
 )
 
 type RestoreManager interface {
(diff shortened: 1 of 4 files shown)


korotkov-aerospike:
Automation passed

pkg/service/restore_manager.go
```diff
@@ -27,3 +35,151 @@ type RestoreManager interface {
...
+	ctx context.Context, request *model.RestoreTimestampRequest,
+) (model.RestoreJobID, error) {
+	return r.timeRunner.RestoreByTime(ctx, request)
+}
+
+// JobStatus returns the status of the job with the given id.
+func (r *RestoreManagerImpl) JobStatus(jobID model.RestoreJobID) (*model.RestoreJobStatus, error) {
+	job, err := r.restoreJobs.getJob(jobID)
+	if err != nil {
+		return nil, err
```
filkeith, line 102: You don't wrap errors. Are these errors so clear that you immediately understand what and where they failed?

pkg/service/restore_manager.go
```diff
@@ -27,3 +35,151 @@ type RestoreManager interface {
...
+	}
+
+	return job.buildStatus(), nil
+}
+
+// CancelRestore cancels an ongoing restore.
+func (r *RestoreManagerImpl) CancelRestore(jobID model.RestoreJobID) error {
+	job, err := r.restoreJobs.getJob(jobID)
+	if err != nil {
+		return err
```
filkeith, line 112: And here, unwrapped error

filkeith (commented):
Consider wrap errors in changed files, not just return `err`

pkg/service/restore_manager.go
korotkov-aerospike: getJob can only retuen NewErrJobNotFound(id) which is pretty self-explanatory

pkg/service/restore_manager.go
korotkov-aerospike: same

filkeith (approved):

PR #518, opened 2026-02-22
BKRS-88 Add restore validation

korotkov-aerospike:


pkg/dto/restore_request.go
```diff
@@ -128,11 +128,11 @@ func (r *RestoreTimestampRequest) ToModel(config *model.Config) (*model.RestoreT
 	}
 
 	return &model.RestoreTimestampRequest{
-		DestinationCluster: cluster,
-		Policy:             r.Policy.ToModel(),
+		DestinationCluster: *cluster,
```
filkeith, line 131: Better change ToModel() functions, so they return exact value, not the pointer. In future it can cause a trouble if you forgot about * dereference and nill pointer error will occure

pkg/dto/restore_request.go
```diff
@@ -128,11 +128,11 @@ func (r *RestoreTimestampRequest) ToModel(config *model.Config) (*model.RestoreT
...
+		Policy:             *r.Policy.ToModel(),
```
filkeith, line 132: Same here, let ToModel() return Policy without &

pkg/model/aerospike_cluster.go
```diff
@@ -76,6 +76,29 @@ func (c *AerospikeCluster) Hash() string {
 	return hex.EncodeToString(hasher.Sum(nil))
 }
 
+// ToString returns a user-friendly cluster identifier.
+// It prefers ClusterLabel; if missing, it uses a stable first seed node string.
+func (c *AerospikeCluster) ToString() string {
+	if c == nil {
+		return nilString
```
filkeith: What is it? Why not just return nil?

pkg/service/restore_validator.go
```diff
@@ -0,0 +1,252 @@
...
+
+	if err := validateDestinationNamespaces(ctx, destinationNamespaces, infoGetter); err != nil {
+		return err
+	}
+
+	return r.checkRunningBackupsConflict(request.DestinationCluster, destinationNamespaces)
+}
+
+// ValidateTimestamp validates point-in-time restore preconditions.
+func (r *restoreValidatorImpl) ValidateTimestamp(
```
filkeith: Same here, no need to be exported

pkg/service/restore_validator.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// validateBackupsEncryption validates that the backups encryption matches the provided policy.
+func validateBackupsEncryption(backups []model.BackupDetails, policy *model.EncryptionPolicy) error {
+	for _, b := range backups {
+		if b.Encryption == "" || b.Encryption == model.EncryptNone {
+			continue
+		}
+		if policy == nil {
+			return fmt.Errorf("b is encrypted with mode '%s', "+
```
filkeith: make error more human, replace b with backup

pkg/service/restore_validator.go
```diff
@@ -0,0 +1,252 @@
...
+				"but no encryption policy was provided in the restore request", b.Encryption)
+		}
+
+		if policy.Mode != b.Encryption {
+			return fmt.Errorf("b is encrypted with mode '%s', "+
```
filkeith: replace b with backup

pkg/service/restore_validator.go
```diff
@@ -0,0 +1,252 @@
...
+				"but the provided encryption policy specifies mode '%s'", b.Encryption, policy.Mode)
+		}
+		if policy.KeyFile == nil &&
+			policy.KeyEnv == nil &&
+			policy.KeySecret == nil {
+			return errors.New("b is encrypted, " +
```
filkeith: replace b with backup

pkg/service/restore_validator.go
```diff
@@ -0,0 +1,252 @@
...
+	infoGetter backup.InfoGetter,
+	backupsByNamespace map[string][]model.BackupDetails,
+) error {
+	for _, backups := range backupsByNamespace {
+		if err := validateBackupsEncryption(backups, request.Policy.EncryptionPolicy); err != nil {
+			return err
+		}
+	}
+
+	sourceNamespaces := sourceNamespacesFromBackupsByNamespace(backupsByNamespace)
```
filkeith: Inside `sourceNamespacesFromBackupsByNamespace` you itterate again through backupsByNamespace. You can collect sourceNamespaces inside loop for _, backups := range backupsByNamespace above

pkg/service/restore_validator.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// sourceNamespacesFromBackups extracts source namespaces from backup metadata entries.
+func sourceNamespacesFromBackups(backups []model.BackupDetails) []string {
+	namespaces := make([]string, 0, len(backups))
+	for _, b := range backups {
+		namespaces = append(namespaces, b.Namespace)
+	}
+
+	return namespaces
```
filkeith, line 211: it wont contain duplicates? May be add slices.Compact() to avoid it

pkg/service/restore_validator.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+		return backupNamespaces[0]
+	}
+	// Empty backup namespace list means routine backs up the whole cluster.
+	if len(backupNamespaces) == 0 {
+		return restoreNamespaces[0]
+	}
+
+	for _, ns := range backupNamespaces {
+		if slices.Contains(restoreNamespaces, ns) {
```
filkeith, line 229: Conatins will itterate through restoreNamespaces, so if there will a few namspaces than ok, but in general it would be faster to convert one slice to map. So it will be o(1). But not crititcal

pkg/service/restore_validator.go
```diff
@@ -0,0 +1,252 @@
...
+	"fmt"
+	"maps"
+	"slices"
+
+	"github.com/aerospike/aerospike-backup-service/v3/pkg/model"
+	"github.com/aerospike/backup-go"
+)
+
+// RestoreValidator validates restore preconditions before actual execution starts.
+type RestoreValidator interface {
```
filkeith, line 15: I know you like to do like that, but Go way is not to store interface and its implementation in one file. Like you define interface where you ned it (near RestoreManager) and then implement it

filkeith (changes requested):

pkg/service/restore_validator.go
korotkov-aerospike: we have max 31 ns, performance should not be a problem :)

pkg/service/restore_validator.go
korotkov-aerospike: its a same package already. Wanna move to sepatate file?

filkeith: checked that you do it in several places, so it's your codestyle. Leave it for now

pkg/service/restore_validator.go
korotkov-aerospike: no, it's logically different checks, should not mix them

pkg/service/restore_validator.go
korotkov-aerospike: should not... but even if they do, we dont care.

// Compact replaces consecutive runs of equal elements with a single copy.
So, it will not help, we should do sort and then compact... I don't see any value in it.
Anyway, when we do check we return any one overlapping namespace, not all of them

filkeith: ofcourse sort then compact, it was just a proposal in general, not step by step implementation guide

pkg/model/aerospike_cluster.go
korotkov-aerospike: inilined.

pkg/dto/restore_request.go
korotkov-aerospike: ToModel() can return nil, it's valid. In this case we use routone's cluster

filkeith (dismissed):

filkeith (approved):

PR #523, opened 2026-03-01
BRKS-107 Incorrect routine invalidation

korotkov-aerospike:
Re-implemented routine invalidaiton in config

### DIFF
--- a/pkg/service/backup_scheduler.go
+++ b/pkg/service/backup_scheduler.go
@@ -87,9 +87,6 @@ func scheduleRoutines(
 	components *BackupComponents,
 	pathService PathService,
 ) error {
-	// Accumulate newly prepared job details and publish them to jobStore only at
-	// the end. This avoids exposing half-built ad-hoc sources while iterating.
-	newJobs := map[string]*quartz.JobDetail{}
 	var errs error
 
 	for _, routine := range routines {
@@ -99,7 +96,7 @@ func scheduleRoutines(
 			continue
 		}
 
-		routine = routine.Copy() // orchestrator will work with it's own copy
+		routine = routine.Copy() // orchestrator will work with its own copy
 		runner := newOrchestrator(routine, components, pathService)
 		// schedule a full backup job for the routine
 		job, err := scheduleFullBackup(scheduler, runner, routine.IntervalCron, routineName)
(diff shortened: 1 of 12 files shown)


pkg/service/backup_scheduler.go
```diff
@@ -80,39 +80,41 @@ func NewScheduler(ctx context.Context, appLogger *slog.Logger) (quartz.Scheduler
...
+	routines []*model.BackupRoutine,
+	components *BackupComponents,
+	pathService PathService,
 ) error {
-	newJobs := map[string]*quartz.JobDetail{}
 	var errs error
-	for routineName, routine := range config.Routines() {
+
+	for _, routine := range routines {
+		routineName := routine.Name
```
filkeith: as I remember we don't ned to copy the value in clousure. And routine.Name is just a field, it's not calculated, so you can use it as is

pkg/service/backup_scheduler.go
```diff
@@ -80,39 +80,41 @@ func NewScheduler(ctx context.Context, appLogger *slog.Logger) (quartz.Scheduler
...
 		if routine.Disabled {
 			slog.Debug("Skipping disabled routine", attr.Routine(routineName))
 			continue
 		}
 
-		routine = routine.Copy() // orchestrator will work with it's own copy
+		routine = routine.Copy() // orchestrator will work with its own copy
```
filkeith: I'm not sure that modifying a (itterated) variable from a range is a good idea. Like I understand that it will contain the same copy, but it looks dangerous. Better introduce new var for that

pkg/service/config_applier.go
```diff
@@ -46,36 +45,63 @@ func (a *DefaultConfigApplier) ApplyNewConfig(ctx context.Context) error {
...
 	}
 
-	slog.Info("Delete scheduled jobs", slog.Any("keys", keys))
-	for _, key := range keys {
-		err = a.scheduler.DeleteJob(key)
-		if err != nil {
-			return fmt.Errorf("failed to delete job %q: %w", key, err)
+	slog.Info("Delete scheduled jobs", slog.Any("keys", keysToDelete))
+	for _, key := range keysToDelete {
+		_ = a.scheduler.DeleteJob(key) // ignore errors because we delete all jobs
```
filkeith, line 93: What will happen if the job fails to delete? It will be running but we won't know about it? May be better log this error, just in case

filkeith (changes requested):

pkg/service/config_applier.go
korotkov-aerospike: it's a 404 error, its fine.

filkeith (approved):
lgtm

PR #525, opened 2026-03-02
BRKS-108 Log restore results

korotkov-aerospike:


pkg/service/jobs_holder.go
```diff
@@ -78,7 +80,7 @@ func (j *restoreJob) addTotalRecords(t uint64) {
 }
 
 // finish marks the job as finished with a given status and error.
-func (j *restoreJob) finish(err error) {
+func (j *restoreJob) finish(err error, logger *slog.Logger) {
```
filkeith, line 83: Have you id model.RestoreJobID in finishJob() may be add it to the log message? So you know which job exactly finished, restored and so on.

filkeith (changes requested):

pkg/service/jobs_holder.go
korotkov-aerospike: Exactly!
Thats why I pass the logger, it was created with job id
logger := slog.With(slog.Any("jobId", jobID))

filkeith: Oh sorry missed that

filkeith (approved):
lgtm

PR #526, opened 2026-03-02
BRKS-109 Add type to last backup metric

korotkov-aerospike:


pkg/service/routine_tracker.go
```diff
@@ -64,65 +64,55 @@ func (t *routineTracker) getState(timeout time.Duration) (*trackerSnapshot, erro
 	defer t.mu.RUnlock()
 
 	snapshot := &trackerSnapshot{
-		full:    currentBackupStatus(t.fullHandler),
-		incr:    currentBackupStatus(t.incrHandler),
+		full:    currentBackupStatus(t.handlers[jobTypeFull]),
```
filkeith, line 67: Do you really need these two separate fields: full and incr? When you have t.handlers and can get its status from there?

korotkov-aerospike: yes, because currentBackupStatus() func actually read from handlers and saves current snapshot

filkeith (approved):

PR #530, opened 2026-03-08
BRKS-118 Rename all var names Cancelled -> Canceled

korotkov-aerospike:
Turns out, linter only checks comments and strings;
Var names are important because they are used to build openAPI spec


filkeith (approved):
lgtm.

As we discussed, 'cancellation` leave as is

PR #531, opened 2026-03-09
BKRS-127 Split restore jobs metric by type

korotkov-aerospike:


pkg/model/restore_result.go
```diff
@@ -1,24 +1,58 @@
...
+)
+
+type RestoreState string
 
 const (
-	JobStatusRunning  JobStatus = "Running"
-	JobStatusDone     JobStatus = "Done"
-	JobStatusFailed   JobStatus = "Failed"
-	JobStatusCanceled JobStatus = "Canceled"
+	RestoreRunning  RestoreState = "running"
```
filkeith: Some of these states will be the same for backup. Maybe you can as it was JobStates, add all possible states to it. And restore will use half of them and backup will use another half, with some overlap. In general we can abstract from operation.

pkg/service/backup_reporter.go
```diff
@@ -0,0 +1,79 @@
...
+	startTime time.Time,
+	duration time.Duration,
+	err error,
+	logger *slog.Logger,
+) {
+	operation := string(backupType) + " backup"
+
+	switch {
+	case err == nil:
+		logger.Debug(operation+" finished", slog.Duration("duration", duration))
```
filkeith, line 51: you have this statuses in constants already. Maybe use them, if smth changed in the future you won't need to update this messages

filkeith (approved):
lgtm

PR #536, opened 2026-03-22
BKRS-150 Refactor backup runner

korotkov-aerospike:


pkg/service/backend_filter.go
```diff
@@ -23,7 +23,7 @@ type BaseFilter struct {
 type RoutineFilter struct {
 	BaseFilter
 	routine *model.BackupRoutine
-	JobType jobType
+	JobType model.BackupType
```
filkeith: You need this field exported?

pkg/service/backup_completion.go
```diff
@@ -91,12 +97,12 @@ func (h *backupCompletionHandler) OnFailure(
 // It also emits the corresponding backup event prometheus metrics.
 func reportBackupOutcome(
 	routineName string,
-	backupType jobType,
+	backupType model.BackupType,
```
filkeith, line 100: Why everywhere you called it jobType but here backupType, choose one

korotkov-aerospike: it will be backupType

filkeith (approved):

PR #543, opened 2026-04-27
Add pprof profiler support for debug and release builds

davi17g:
@filkeith Can we create similar PR for `absctl`?

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -155,7 +155,7 @@ jobs:
             exit 1
           fi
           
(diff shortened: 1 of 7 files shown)


korotkov-aerospike (dismissed):

filkeith:
In general, yes of course, but it can be tricky to catch the moment while cli tool is running to send http request to get the pprof. I'm usually automatically dumping profiling to the file every 10 seconds and then use it

korotkov-aerospike (approved):

PR #546, opened 2026-05-06
BRKS-185 Improve backup progress accuracy

korotkov-aerospike:
Was: estimates calculate percent, progress = percent / 100
Now: estimates calculate ration, percent = ration * 100

Don't loose accuacy.


pkg/dto/routine_state.go
```diff
@@ -77,7 +77,7 @@ func (r *RunningJob) fromModel(m *model.RunningJob) {
 	r.DoneRecords = m.DoneRecords
 	r.StartTime = m.StartTime
 	r.FinishTime = m.FinishTime
-	r.PercentageDone = m.PercentageDone
+	r.PercentageDone = uint(m.Progress * 100)
```
filkeith, line 80: Be ready that uint() will just cut zeroes, without rounding.

filkeith: Also r.PercentageDone can be uint8, as 100 is the maximum val.

filkeith (approved):

pkg/dto/routine_state.go
korotkov-aerospike: yeah, that's fine

PR #552, opened 2026-05-12
BKRD-187 Change chunk size default value

filkeith:
- added validation for file limit and chunk size
- limit increased


pkg/dto/backup_routine.go
```diff
@@ -265,6 +269,43 @@ func (r *BackupRoutine) ToModel(config *model.BackupConfig, name string) (*model
 	}, nil
 }
 
+func validateFileLimit(policy *model.BackupPolicy, storage model.Storage) error {
+	if storage == nil || policy == nil {
+		return nil
+	}
+
+	// Skip this check for the local storage.
+	if strings.Contains(storage.String(), "LocalStorage") {
```
korotkov-aerospike: isnt it better to try typecast?

pkg/dto/backup_routine.go
```diff
@@ -265,6 +269,43 @@ func (r *BackupRoutine) ToModel(config *model.BackupConfig, name string) (*model
...
+		return nil
+	}
+
+	// It is easier to set default values than do a bunch of comparisons.
+	fileLimit := 250 * 1024 * 1024
```
korotkov-aerospike: We have func 

(p *BackupPolicy) GetFileLimitOrDefault() int {
	if p.FileLimit != nil {
		return *p.FileLimit
	}

	return *defaultConfig.backupPolicy.FileLimit
}

pkg/dto/backup_routine.go
```diff
@@ -265,6 +269,43 @@ func (r *BackupRoutine) ToModel(config *model.BackupConfig, name string) (*model
...
+	if policy.FileLimit != nil {
+		fileLimit = *policy.FileLimit * 1024 * 1024
+	}
+
+	partSize := 50 * 1024 * 1024
+	if storage.GetPartSize() != nil {
+		partSize = *storage.GetPartSize()
+	}
+
+	const maxChunks = 10_000
```
korotkov-aerospike: extract const from function; It's same for all storages?

korotkov-aerospike (changes requested):

pkg/dto/backup_routine.go
filkeith: moved

pkg/dto/backup_routine.go
filkeith: thx, fixed!

pkg/dto/backup_routine.go
filkeith: fixed

korotkov-aerospike (approved):

PR #558, opened 2026-05-24
BKRS-194 Add connectivity checks on startup

korotkov-aerospike:
Extends S3, GCP, and Azure storage connectivity checks at client initialization so misconfigured credentials or IAM policies surface early instead of at  backup time. After confirming bucket/container access, each backend verifies read permissions (S3/Azure via a minimal list call; GCP via TestPermissions  for storage.objects.get and storage.objects.list), failing startup if those are missing. Write and delete capabilities are checked non-blockingly—S3 and  Azure via a small probe object (.abs-connectivity-check, with S3 using a multipart upload path aligned with real backup writes; GCP via IAM permission  tests)—logging warnings when upload or delete is denied so read-only or least-privilege setups can still start while alerting operators that backups may  fail later.


pkg/service/storage/azure.go
```diff
@@ -182,5 +184,30 @@ func checkAzureConnectivity(ctx context.Context, client *azblob.Client, containe
...
+	}
+
+	blob := cc.NewBlockBlobClient(connectivityProbeKey)
+	_, err = blob.UploadBuffer(ctx, []byte{}, nil)
+	if err != nil {
+		slog.Warn("azure blob storage upload permission check failed; backup writes may fail at runtime",
+			slog.String("container", container),
+			attr.Error(err),
+		)
+		return nil
```
filkeith, line 201: Are you sure that the upload permission error should be ignored? Service will start and no one will check logs foe warning

pkg/service/storage/azure.go
```diff
@@ -182,5 +184,30 @@ func checkAzureConnectivity(ctx context.Context, client *azblob.Client, containe
...
+	}
+
+	_, err = blob.Delete(ctx, nil)
+	if err != nil && !bloberror.HasCode(err, bloberror.BlobNotFound) {
+		slog.Warn("azure blob storage delete permission check failed; backup writes or cleanup may fail at runtime",
```
filkeith, line 206: same here, we should return a validation error?

pkg/service/storage/azure.go
```diff
@@ -182,5 +184,30 @@ func checkAzureConnectivity(ctx context.Context, client *azblob.Client, containe
...
+	if err != nil {
+		slog.Warn("azure blob storage upload permission check failed; backup writes may fail at runtime",
+			slog.String("container", container),
+			attr.Error(err),
+		)
+		return nil
+	}
+
+	_, err = blob.Delete(ctx, nil)
+	if err != nil && !bloberror.HasCode(err, bloberror.BlobNotFound) {
```
filkeith, line 205: If deletion failed, your test object will stay in the cloud, may be you should print to logs smth like: remove file manually... 
After bunch of restarts you can flood cloud with test objects

pkg/service/storage/gcp.go
```diff
@@ -133,10 +136,35 @@ func checkGcpConnectivity(ctx context.Context, client gcp.Client, bucket string)
...
+	w := obj.NewWriter(ctx)
+	if err = w.Close(); err != nil {
+		slog.Warn("gcp storage upload permission check failed; backup writes may fail at runtime",
+			slog.String("bucket", bucket),
+			attr.Error(err),
+		)
+		return nil
+	}
+
+	if err = obj.Delete(ctx); err != nil {
```
filkeith: for GCP we can use client.Bucket(bucket).IAM().TestPermissions(ctx, required)
to check permissions without creating anything

pkg/service/storage/gcp.go
```diff
@@ -133,10 +136,35 @@ func checkGcpConnectivity(ctx context.Context, client gcp.Client, bucket string)
...
+		slog.Warn("gcp storage delete permission check failed; backup writes or cleanup may fail at runtime",
```
filkeith, line 177: same here, we just warn on deletion fail and leave tst file in the cloud

pkg/service/storage/s3.go
```diff
@@ -175,10 +178,42 @@ func checkS3Connectivity(ctx context.Context, client *awsS3.Client, bucket strin
...
 
+	_, err = client.ListObjectsV2(ctx, &awsS3.ListObjectsV2Input{
+		Bucket:  aws.String(bucket),
+		MaxKeys: aws.Int32(1),
+	})
+	if err != nil {
+		return fmt.Errorf("s3 storage read permission check failed: %w", err)
+	}
+
+	_, err = client.PutObject(ctx, &awsS3.PutObjectInput{
```
filkeith: we use multipart upload, just for checking start/complete upload PutObject will be enough, but if we want to cover all checks, we should check AbortMultipartUpload - as it is used in our Go lib, and it is separate permission

filkeith (changes requested):

pkg/service/storage/azure.go
korotkov-aerospike: it's the same file, "connectivityProbeKey", so it should be ok.

pkg/service/storage/gcp.go
korotkov-aerospike: Agree, done.

filkeith (approved):
lgtm

PR #564, opened 2026-06-07
BKRS-212 Make routine optional

korotkov-aerospike:
## Description
  Enhances the **Restore-by-Timestamp** API to support cross-region restores without requiring a locally configured backup routine. 
  Previously, customers performing cross-region migrations (Backup → Copy → Restore) could not use Restore-by-Timestamp unless the referenced backup routine  existed in the local ABS configuration. This PR introduces optional runtime overrides for storage, destination cluster, and secret agent, allowing the API to bypass local routine validation when explicit configurations are provided.
  ### Key Changes
  * **Runtime Overrides:** Added `source` / `source-name` to `RestoreTimestampRequest`. The `destination`, `source`, and `secret-agent` fields now act as   optional field-by-field overrides.
  * **Smart Fallback:** If an override is omitted, the system falls back to resolving the configuration from the referenced `routineName`.
  * **Model Refactoring:** Replaced the internal `BackupRoutine` object in `model.RestoreTimestampRequest` with explicit `RoutineName` and `Storage` fields   to avoid constructing "fake" routine objects.
  * **Validation & Resolution:** Updated `Validate()` to allow empty values for optional overrides and refactored `ToModel()` to handle the request-first,   routine-fallback resolution logic.
  * **Backend Updates:** Updated `RoutineFilter` and the time restore runner to operate directly on `routineName` and `storage`.
  * **Documentation & Tests:** Added comprehensive unit tests for the new override scenarios and updated OpenAPI/Markdown documentation to reflect the new   behavior.


filkeith (approved):
lgtm

PR #570, opened 2026-07-02
BKRS-238 Fair namespace backup

korotkov-aerospike:


pkg/util/syncutil/random_semaphore.go
```diff
@@ -0,0 +1,157 @@
...
+// RandomSemaphore is a counting semaphore that wakes blocked waiters in random
+// order rather than FIFO. Random selection spreads wakeups across competing
+// goroutines, which helps avoid convoying when many waiters queue for the same
+// tokens (for example, scan parallelism slots shared across namespaces).
+//
+// Acquire takes available tokens immediately when s.tokens >= n, even if other
+// goroutines are already waiting.
+type RandomSemaphore struct {
+	mu      sync.Mutex
+	tokens  int64
```
filkeith: Are u sure that 1 field tokens will be enough to track everything? Like if one consumer fails to return token, we will lost it forever. May be we should save smth like initial Capacity, and compare with it?

pkg/util/syncutil/random_semaphore.go
```diff
@@ -0,0 +1,157 @@
...
+		return nil
+	}
+}
+
+// Release returns n tokens to the semaphore and wakes waiters that can proceed.
+func (s *RandomSemaphore) Release(n int64) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.tokens += n
```
filkeith: Here we can release over Capacity

pkg/util/syncutil/random_semaphore.go
```diff
@@ -0,0 +1,157 @@
...
+	if n == 0 || s.tokens == 0 {
+		return
+	}
+
+	rand.Shuffle(n, func(i, j int) {
+		s.waiters[i], s.waiters[j] = s.waiters[j], s.waiters[i]
+	})
+
+	alive := s.waiters[:0]
+	for _, w := range s.waiters[:n] {
```
filkeith: s.waiters[:n] looks like overhead, like how much waiters there will be 16? 128?We can just range s.waiters, but ok it is not a problem.

filkeith (changes requested):

filkeith (dismissed):
lgtm

filkeith (approved):

PR #572, opened 2026-07-06
Rescan storage after backup 

korotkov-aerospike:
Instead of setting last timestamp, we trigger new storage rescan. That make logic more robust when multithreading

### DIFF
--- a/pkg/service/routine_tracker.go
+++ b/pkg/service/routine_tracker.go
@@ -140,7 +140,7 @@ func (t *routineTracker) cancelScan() {
 }
 
 // beginScan prepares the tracker for a new storage scan.
-// It unblocks any getState callers waiting on a previous (now-cancelled) scan
+// It unblocks any getState callers waiting on a previous (now-canceled) scan
 // and installs a fresh open channel for the new scan.
 // Returns the channel that the caller must pass to endScan when done.
 func (t *routineTracker) beginScan() chan struct{} {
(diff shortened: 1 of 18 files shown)


pkg/service/routine_tracker.go
```diff
@@ -50,24 +47,36 @@ type trackerSnapshot struct {
...
-	case <-t.initialSyncDone:
-		// Sync is done, proceed
-	case <-time.After(timeout):
-		return nil, context.DeadlineExceeded
+	deadline := time.After(timeout)
+
+	for {
+		t.mu.RLock()
+		done := t.scanDone
+		t.mu.RUnlock()
```
filkeith, line 58: You are rewriting the variable, are you sure that Read Lock is enough?

pkg/service/routine_tracker.go
```diff
@@ -50,24 +47,36 @@ type trackerSnapshot struct {
...
+
+		select {
+		case <-done:
+		case <-deadline:
+			return nil, context.DeadlineExceeded
+		}
+
+		// If a new scan started while we were waiting (beginScan closes the
+		// old channel and installs a new one), loop to wait for the new scan.
+		t.mu.RLock()
```
filkeith, line 68: same here, it should be Write lock?

pkg/service/routine_tracker.go
korotkov-aerospike: we are READING t.scanDone (global)
and WRITE to `done` - local var.

2 threads can access status at the same time

pkg/service/routine_tracker.go
korotkov-aerospike: same, we read t.scanDone

filkeith (dismissed):
If tests pass, ok. But we need to think about simplifying the channels' logic

filkeith (approved):

PR #600, opened 2026-08-09
BKRS-292 Remove the secret resolver cache entirely

korotkov-aerospike:
## What does this change do?

Removes the redundant 24-hour secret resolver cache that used sliding expiry semantics, causing secrets to live indefinitely under regular access patterns.

**Summary:**
- Strip `storageDuration`, `cacheKey`, and `collections` dependency from resolver
- Resolver becomes a direct pass-through to `backup.ParseSecret`
- Add comprehensive tests proving uncached behavior and fail-closed semantics
- Document credential resolution and rotation semantics in configuration guide

The cache is redundant because every call site already sits behind a client cache:
- S3/GCP clients cache for process lifetime.
- Azure clients cache with one-hour sliding TTL but remain effectively permanent under steady load
- Cluster passwords re-resolve on idle client recreation (~10-second default)

This change removes an internal caching layer only; it does not enable live storage credential rotation, as the real binding constraint is the per-storage-type client caches which bake resolved credentials into the constructed client for their lifetime.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Chore / refactor / CI

## Checklist

- [x] `make pr` passes locally (tidy, mocks, format, lint, tests, OpenAPI, README generation).
- [x] Generated artifacts are committed and up to date: no API/OpenAPI changes; internal resolver only.
- [x] Configuration documentation updated to explain credential rotation semantics.

## Additional context

Fixes BKRS-292. The resolver tests use TCP-based Secret Agent protocol simulation to verify:
1. **Uncached behavior**: Repeated resolutions make independent connections to the agent
2. **Fail-closed**: Resolution errors propagate without stale-secret fallback
3. **Pass-through**: Literal values and nil agents bypass the agent layer

All existing tests pass; no regression in backup/restore for S3, GCP, Azure, or cluster paths.

### DIFF
--- a/internal/app/bootstrap.go
+++ b/internal/app/bootstrap.go
@@ -32,7 +32,7 @@ func InitComponents(
 	configFile string,
 	remote bool,
 ) (quartz.Scheduler, *handlers.Service, error) {
-	resolver := secrets.NewResolver(ctx)
+	resolver := secrets.NewResolver()
 	operations := newStorageOperations(ctx, resolver)
 	clientManager, nsValidator := newAerospikeLayer(resolver)
 
--- a/internal/app/bootstrap.go
+++ b/internal/app/bootstrap.go
@@ -33,7 +33,7 @@ func InitComponents(
 	remote bool,
 ) (quartz.Scheduler, *handlers.Service, error) {
 	resolver := secrets.NewResolver()
-	operations := newStorageOperations(ctx, resolver)
+	operations := newStorageOperations(resolver)
 	clientManager, nsValidator := newAerospikeLayer(resolver)
 
 	config, configurationManager, err := configuration.Load(ctx, configFile, remote, nsValidator, operations)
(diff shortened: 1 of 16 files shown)


internal/app/bootstrap.go
```diff
@@ -32,8 +32,8 @@ func InitComponents(
 	configFile string,
 	remote bool,
 ) (quartz.Scheduler, *handlers.Service, error) {
-	resolver := secrets.NewResolver(ctx)
-	operations := newStorageOperations(ctx, resolver)
+	resolver := secrets.NewResolver()
```
filkeith, line 35: Look like now we don't need resolver wrapper at all. It can be just:
resolver := func(value string) (string, error) {
  return backup.ParseSecret(ctx, agentConfig, value)
}

filkeith (approved):
looks goos, but we can remove resolver at all

internal/app/bootstrap.go
korotkov-aerospike: let's keep it for uniformality

PR #604, opened 2026-08-11
BKRS-272 Rate limiter hardening

korotkov-aerospike:
## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Chore / refactor / CI

## Checklist

- [ ] `make pr` passes locally (tidy, mocks, format, lint, tests, OpenAPI, README generation).
- [ ] Tests were added or updated for the change.
- [ ] Generated artifacts are committed and up to date (`make generated-check` passes): mocks, OpenAPI spec, README/docs.


internal/server/middleware/rate_limiter.go
```diff
@@ -78,76 +91,137 @@ func newIPWhiteList(ipList []string) *IPWhiteList {
...
 	tokenBucketSize int
+	now             func() time.Time
+	idleTTL         time.Duration
+	cleanupTicker   ticker
 }
 
 // NewIPRateLimiter returns a new IPRateLimiter.
-func NewIPRateLimiter(tps rate.Limit, size int) *IPRateLimiter {
+func NewIPRateLimiter(ctx context.Context, tps rate.Limit, size int) *IPRateLimiter {
+	return newIPRateLimiter(
```
filkeith: maybe add a comment, that whis wrapper is used for tests. Because without it, it looks very strange

internal/server/middleware/rate_limiter.go
```diff
@@ -78,76 +91,137 @@ func newIPWhiteList(ipList []string) *IPWhiteList {
...
 	ipLimiter.Lock()
-	limiter, exists := ipLimiter.limiters[IPAddress(ipAddr)]
+	defer ipLimiter.Unlock()
 
-	if !exists {
-		ipLimiter.Unlock()
-		return ipLimiter.AddLimiter(ipAddr)
+	for ip, entry := range ipLimiter.limiters {
+		if now.Sub(entry.lastSeen) >= ipLimiter.idleTTL {
+			delete(ipLimiter.limiters, ip)
```
filkeith, line 186: I'm not sure that delete() returns memory. If we saved 1000 ips, after this operation map wil be allocated for 1000 elements. I advise using LRU list (or existing LRU library). Also, during this clean we hold mutex lock, so no new element will be added.  So in normal conditions it won't affect us. But if someopne will try to flood ABS with requests, and there will be 100k elements, api queries became slow because it will wait while the list will be cleaned.

filkeith (changes requested):

internal/server/middleware/rate_limiter.go
korotkov-aerospike: yeah, you're technically right, but I don't think we're ever hit even 1% of that traffic :)

filkeith: will know when someone tries to break us

filkeith (approved):

PR #610, opened 2026-08-13
Refactor release workflows and update versioning metadata

davi17g:


.github/workflows/release.yml
```diff
@@ -1,259 +1,105 @@
-name: Promote Images to DockerHub
+name: Promote Release Bundle to PROD
```
arrowplum: making a release should != promoting it to prod. making a release should just be making a release and placing on jfrog. from there it should be promoted to test, tests should run using that artifact and if pass it should be promoted to stage.

At this point further promotion to something public needs to be done by either a PM or EM if there is no PM for that project.

https://aerospike.atlassian.net/wiki/spaces/DevOps/pages/4577493063/Pipeline+with+gates+QE+Cloud+Product is the docs

build/docker-compose/docker-compose.yaml
```diff
@@ -47,7 +47,7 @@ services:
       retries: 3
 
   aerospike-backup-service:
-    image: aerospike/aerospike-backup-service:3.6.1
+    image: aerospike/aerospike-backup-service:0.0.0-test-dgerchikov.1
```
arrowplum: I think you are just testing but t test something that has been deployed to jfrog just use the image for that jfrog repository here.

.github/workflows/pre-release.yml
```diff
@@ -81,594 +85,367 @@ jobs:
...
+      # build-packages/deploy-packages and build-helm/deploy-helm were given as jf-build-id.
+      # The container build's build-NUMBER is github.run_number (the setup-jfrog-cli default),
+      # since reusable_docker-build-deploy.yaml has no jf-build-id input at all.
+      jf-build-names: aerospike-backup-service-packages:${{ needs.prepare.outputs.packages_build_id }},aerospike-backup-service-helm:${{ needs.prepare.outputs.helm_build_id }},aerospike-backup-service-container:${{ github.run_number }}
+
+  # ---------------------------------------------------------------------------
+  # GitHub Release assets (no shared-workflow equivalent)
+  # ---------------------------------------------------------------------------
+
+  publish-release-assets:
```
arrowplum: This job shouldn't be attached to the same flow here (because github assets mean the release has already been made public or completed therefore it needs to have gone through  all stages and gates so it wouldn't be immediately post-able. It should also be immutable so we aren't changing releases under people's feet.

If we break this out into a separate action it should take a bundle from jfrog and upload its components as a release. All checksums and gpg would have already been done and are on jfrog which is the source of truth here. Nothing should be regenerated like is done in this job.

arrowplum (changes requested):

davi17g:
@arrowplum Following up on your review — here's how the release flow will work with these changes (same as documented in `docs/development.md`, mirrored in `absctl`'s README in [absctl#223](https://github.com/aerospike/absctl/pull/223)):

1. We create a new tag and push it - it triggers `pre-release.yml`. Pre-release does the following:
   1. Builds all the artifacts (Docker images, rpm/deb packages, Helm chart - for ABS)
   2. Signs them and deploys everything to JFrog `DEV`
   3. Creates a unified release bundle
   4. Automatically transitions the release bundle from `DEV` to `TEST`
2. QA/devs test the artifacts by downloading them from `TEST` in JFrog; once tests pass, they transition the release bundle to `STAGE` (manually, via the JFrog UI - not automated)
3. A PM/EM reviews and transitions the release bundle from `STAGE` to `PROD` via the JFrog UI (this is the gate that makes the release public; we don't promote to PROD ourselves in CI anymore)
4. Once the bundle is on `PROD`:
   1. Docker Hub mirroring happens automatically and externally (existing JFrog webhook → `artifact-publisher`) - no action needed from us
   2. Someone with repo access (not necessarily the same PM/EM from step 3) manually triggers `release.yml` (`workflow_dispatch`, version input), which:
      1. Verifies the bundle has actually been promoted to PROD in JFrog (fails with a clear error otherwise)
      2. Downloads the already-signed rpm/deb/checksums (and Helm chart, for ABS) straight from JFrog - nothing is regenerated or re-signed
      3. Publishes them as a new, immutable GitHub Release (created directly as published, not draft/pre-release)

A few questions:

1. Can you please confirm the flow?
2. What about the release changelog - should it be generated automatically using something from `shared-workflows` (similar to `release-changelog.yaml`'s conventional-changelog approach), or is it fine that it stays manually maintained (`CHANGELOG.md` updated by devs per PR, like today)?
3. Can you please confirm that `artifact-publisher` will work with both `absctl` and `aerospike-backup-service`? I noticed `citrusleaf/artifact-publisher`'s `.github/config/publish-allow-list.json` still has an entry for `{project_key: ecosystem, bundle_name: aerospike-backup-service}`, which looks stale now that we're on the `database` project, and there's no entry at all yet for `absctl`.

build/docker-compose/docker-compose.yaml
davi17g: Yes, that happened by mistake, thanks for pointing

arrowplum:
>2.  QA/devs test the artifacts by downloading them from TEST in JFrog; once tests pass, they transition the release bundle to STAGE (manually, via the JFrog UI - not automated)

This does not have to be manual (it can be of course) if you automate using a workflow it will work with the same oicd parameters.

>4.i Docker Hub mirroring happens automatically and externally (existing JFrog webhook → artifact-publisher) - no action needed from us

Some action items may be needed from @aerospike/dev-infra to update the allowlist

For the questions:

> Can you confirm the flow

Just those above comments

> What about the release changelog - should it be generated automatically using something from shared-workflows (similar to release-changelog.yaml's conventional-changelog approach), or is it fine that it stays manually maintained (CHANGELOG.md updated by devs per PR, like today)?

There is no problem if you do it the same as today but you can use those tools to help.

> Can you please confirm that artifact-publisher will work with both absctl and aerospike-backup-service? I noticed citrusleaf/artifact-publisher's .github/config/publish-allow-list.json still has an entry for {project_key: ecosystem, bundle_name: aerospike-backup-service}, which looks stale now that we're on the database project, and there's no entry at all yet for absctl.

As you saw there might be allowlist updates needed there. The steps to push can be done manually as well if needed.

arrowplum (dismissed):
Thank you for taking care. Changes look good.

korotkov-aerospike (approved):

korotkov-aerospike (dismissed):

korotkov-aerospike (approved):

PR #633, opened 2026-08-23
BKRS-285 Add HTTPS service config and a server-safe TLS builder

korotkov-aerospike:
## What does this change do?

Phase 1 TLS, part 1 of [BKRS-285](https://aerospike.atlassian.net/browse/BKRS-285) / parent [BKRS-263](https://aerospike.atlassian.net/browse/BKRS-263).

Adds a sibling `service.https` config block (not nested under `http`) with validation and a server-safe `*tls.Config` builder. Omitting `https` keeps today's plaintext HTTP listener on 8080. This change does **not** start an HTTPS listener; cert reload and `HTTPServer.Start` wiring are part 2.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Chore / refactor / CI

## Checklist

- [ ] `make pr` passes locally (tidy, mocks, format, lint, tests, OpenAPI, README generation).
- [x] Generated artifacts are committed and up to date (`make generated-check` passes): mocks, OpenAPI spec, README/docs.

## Additional context

- HTTP and HTTPS use `disabled` (default `false`), not `enabled`. Omitted `http` still means enabled; omitted `https` means no HTTPS listener. At least one listener must stay on; if both are on, ports must differ.
- Enabled HTTPS requires matching `cert-file`/`key-file`. `client-auth` other than `none` requires `client-ca-file`.
- `internal/server/tlsconfig` is a new stdlib builder. It does **not** wrap `pkg/tlsconfig` / Aerospike client TLS (those include `tls.InsecureCipherSuites()`). The HTTPS builder uses only `tls.CipherSuites()`; a unit test asserts disjointness from insecure suites.
- Closed config enums (`AuthMode`, `TimestampFormat`, `TLSMinVersion`, `TLSClientAuth`) parse in the DTO layer with typed `ToModel()` switches. The model package no longer parses those config strings.

## Test plan

- [ ] Config with no `service.https` still validates and matches current HTTP-on-8080 defaults.
- [ ] `http.disabled: true` without `https` fails validation.
- [ ] Both listeners enabled on the same port fails validation.
- [ ] Enabled `https` with only one of `cert-file`/`key-file`, unmatched cert/key, unknown `min-version`/`cipher-suites`, or `client-auth != none` without `client-ca-file` fails validation.
- [ ] `HTTPSServerConfig.Compare` and `HTTPServerConfig.Compare` (`disabled`) report diffs; `key-file-password` diffs do not leak the secret.
- [ ] `internal/server/tlsconfig` builder suite set has no overlap with `tls.InsecureCipherSuites()`.
- [ ] Confirm HTTPS is still not bound in `HTTPServer.Start` (expected until part 2).

[BKRS-285]: https://aerospike.atlassian.net/browse/BKRS-285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BKRS-263]: https://aerospike.atlassian.net/browse/BKRS-263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/dto/https_server_config.go
```diff
@@ -0,0 +1,284 @@
...
+	return result
+}()
+
+const clientCAField = "client-ca-file"
+
+// HTTPSServerConfig represents the service's HTTPS server configuration.
+// @Description HTTPSServerConfig represents the service's HTTPS server configuration.
+//
+//nolint:lll
+type HTTPSServerConfig struct {
```
filkeith: HTTPSServerConfig is the same as HTTP but some more fields, why not embed HTTPServerConfig here as SecretAgent?

Also think abut changing the name oif this struct. I noticed the difference between HTTPSServerConfig and HTTPServerConfig only because I see full PR changelog. In other situation it is too hard to detect one more S. What about ServerConfigHTTP and ServerConfigHTTPS or smth else

pkg/dto/http_server_config.go
```diff
@@ -13,6 +13,8 @@ import (
 // HTTPServerConfig represents the service's HTTP server configuration.
 // @Description HTTPServerConfig represents the service's HTTP server configuration.
 type HTTPServerConfig struct {
```
filkeith: In general, do you really need a separate HTTPServerConfig type?  It can be 1 type for both HTTP and HTTPS, HTTP will not contain any tls configuration

filkeith (changes requested):

pkg/dto/http_server_config.go
korotkov-aerospike: maybe; let me think

pkg/dto/https_server_config.go
korotkov-aerospike: cannot inline completely, but extracted common code.

korotkov-aerospike: renamed; that doubled PR size, sorry :)

filkeith (approved):
LGTM

PR #643, opened 2026-08-31
BKRS-370 Timezone for backup schedules

korotkov-aerospike:
## What does this change do?

Adds optional `schedule-timezone` on `service.backup` and on each backup routine so cron expressions can be evaluated in UTC (default), `Local`, or an IANA zone such as `America/New_York`. Backup folder names and `timestamp-format` suffixes stay UTC.

Fixes https://aerospike.atlassian.net/browse/BKRS-370

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Chore / refactor / CI

## Checklist

- [ ] `make pr` passes locally (tidy, mocks, format, lint, tests, docs).

## Additional context

- Omitted `schedule-timezone` keeps today's UTC cron behavior.
- A routine value overrides the service default; an empty routine value inherits it. `fromModel` omits the inherited field so a read-modify-write does not bake it in.
- Invalid values (`EST`, unknown IANA names) fail config validation.
- Scheduler, `next-full` / `next-incremental`, and the same-tick full-versus-incremental skip all use `routine.CronLocation()`.
- IANA names resolve without a `tzdata` package via `import _ "time/tzdata"`.
- `timestamp-format` path suffixes are now always rendered in UTC (folder identity is still epoch milliseconds).


pkg/model/backup_routine.go
```diff
@@ -22,6 +23,13 @@ type BackupRoutine struct {
 	IntervalCron string
 	// The interval for incremental backup as a cron expression string (optional).
 	IncrIntervalCron string
+	// Timezone is the resolved timezone for evaluating cron expressions. Not nil
+	// after DTO conversion. Inherited from service.backup.schedule-timezone when
+	// ConfiguredTimezone is empty.
+	Timezone *time.Location
+	// ConfiguredTimezone is this routine's schedule-timezone as configured.
+	// A blank value means the routine inherits the service default.
+	ConfiguredTimezone string
```
filkeith: why don't you make it *time.Location ? Like you need string only for YAML/JSON unmarshaling, then you can parse it and save as normal timezone

filkeith (dismissed):
LGTM, but check my comment

filkeith (approved):
lgtm

PR #644, opened 2026-09-01
BKRS-376 Reload HTTPS client CA file without restart

korotkov-aerospike:
## What does this change do?

Reloads `service.https.client-ca-file` when the bundle is rewritten at the configured path, allowing mTLS trust rotation without restarting the service. New TLS handshakes use an atomically swapped CA pool; malformed or partially valid replacements retain the last complete pool and are retried.

The rotation tests cover the safe overlap sequence (old CA, old + new CAs, new CA), rejection of old clients after cutover, malformed-bundle ERROR logging, and recovery after a valid replacement.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Chore / refactor / CI

## Checklist

- [x] If this changes the configuration file or REST API in a breaking way, the migration guide is updated. (Not breaking; migration and security documentation are updated.)

## Additional context

Certificate/key watchers remain independent from the client-CA watcher so an invalid CA replacement cannot interfere with server certificate rotation.


internal/server/tlsconfig/reloader.go
```diff
@@ -78,8 +100,20 @@ func (noOpReloader) GetCertificate(*tls.ClientHelloInfo) (*tls.Certificate, erro
...
+func (noOpReloader) ClientCAs() (*x509.CertPool, error) {
+	return nil, errors.New("HTTPS client CA pool is not loaded")
+}
+
+// Load reads the key pair and the client CA pool and serves them to new handshakes.
+func (r *certificateReloader) Load(ctx context.Context) error {
+	if err := r.loadKeyPair(ctx); err != nil {
+		return err
+	}
+
```
filkeith, line 112: can't you potentially catch situation between this 2 loading, when key was loaded from new cert and CA is old? loadKeyPair() and loadClientCAs(). May be you should lock mutex here in Load() function. Like we lock, update everything and then unlock. Not lock each operation

filkeith (changes requested):

filkeith (approved):

PR #651, opened 2026-09-06
BKRS-386 Add nilaway linter

korotkov-aerospike:
## What does this change do?

Makes conversion and model-mutation contracts total so NilAway can reason about them, and adds a `make nilaway` check for production packages.

NilAway treated several `nil` success paths as real panics. Those paths were either impossible (value copies, correlated flags) or represented invalid config that should be rejected at the boundary.

- `BackupConfig.copy` is a value method and cannot return nil.
- `AddCluster`, `AddRoutine`, and `AddStorage` reject nil (and unsupported storage types).
- Model→DTO constructors no longer return nil for valid model state.
- Restore timestamp overrides check `IsEmpty()` before `ToModel()`; destination conversion no longer uses `(nil, nil)` for “omitted”.
- `BackupRoutine.Copy()` no longer has a nil-success path.
- Client cache release uses `targetInfo != nil` as the single match signal.
- `make nilaway` skips tests, generated mocks, and the docs generator.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Chore / refactor / CI

## Checklist

- [ ] `make pr` passes locally (tidy, mocks, format, lint, tests, docs).

## Additional context

`Validate()` remains the DTO shape check. `ToModel()` is conversion and lookup only; it is assumed to run on validated objects. Optional restore-by-timestamp fields are still resolved by the caller via `IsEmpty()`, not by a third `Resolve` step.

NilAway is not a golangci-lint plugin here; CI/local use `go run go.uber.org/nilaway/cmd/nilaway@…` with a pinned pseudo-version (no tagged releases).

### DIFF
--- a/.github/workflows/nilaway.yml
+++ b/.github/workflows/nilaway.yml
@@ -1,0 +1,38 @@
+# Fails if NilAway reports a potential nil panic in production code.
+# Run "make nilaway" locally. Tests, generated mocks, and the docs generator are excluded.
+name: NilAway
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - main
+      - dev
+      - 'v[0-9]+'
+
+permissions:
+  contents: read
+
+jobs:
+  nilaway:
+    name: nilaway
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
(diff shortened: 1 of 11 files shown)


filkeith (approved):
lgtm

PR #659, opened 2026-09-09
BKRS-394 Resolve cluster TLS key-file-password through the Secret Agent

korotkov-aerospike:
## What does this change do?

Fixes [BKRS-394](https://aerospike.atlassian.net/browse/BKRS-394): a cluster whose encrypted client TLS key is protected by a Secret Agent-backed `key-file-password` silently connected **without TLS**.

`clientFactory.setTLSConfig` never resolved `AerospikeCluster.TLS.KeyfilePassword` before using it as the decryption passphrase, and swallowed the resulting error into a `slog.Error` line instead of returning it. For a cluster configured with `key-file-password: secrets:<resource>:<key>` (a documented, DTO-validated configuration):

1. Config validation passed — `Prober.probeCluster` resolves the reference correctly, so the config looked healthy at apply time.
2. Every real connection failed to decrypt the key, because the literal `secrets:...` string was passed to `x509.DecryptPEMBlock` (`x509: decryption password incorrect`).
3. Because that error was only logged, `policy.TlsConfig` stayed `nil` and the Aerospike client fell back to a plaintext connection to seed nodes configured to require TLS.

Root cause was two independent implementations of "build a cluster's TLS config": `probeCluster` resolved the secret, `setTLSConfig` didn't — and had no resolver dependency, so it structurally couldn't.

### Changes

- `setTLSConfig` resolves the password and returns resolution / TLS-config failures instead of logging them, so a security control fails closed.
- New `secrets.ClusterTLSResolver` owns "give me this cluster's TLS config with its secrets resolved", mirroring the existing `secrets.PasswordResolver` shape. Both `probeCluster` and `setTLSConfig` delegate to it in one line, so the agent lookup and the nil-TLS default live in exactly one place and can't drift apart again.
- Clarified `model.TLS.KeyfilePassword`'s doc comment, which said the value "is used verbatim" without noting it may still be an unresolved reference — the assumption that led to the bug.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Chore / refactor / CI

## Checklist

## Additional context

**Why existing tests missed it.** The integration auth matrix covers auth mode x TLS mode thoroughly, and separately covers value source (literal / password-path / Secret Agent) — but only ever for `Credentials.Password`, never for `TLS.KeyfilePassword`. `TestMutualTLS` used a hardcoded literal passphrase, `TestProbeCluster` never set `KeyfilePassword` at all, and `client_factory.go` had no unit tests. The gap was the *intersection* of two axes that were each well covered alone.

**Test coverage added.**
- `TestClusterAuth/TestMutualTLS/internal_with_key-file-password_from_secret_agent` — end-to-end backup over real mTLS with an encrypted client key whose password is resolved through a real Secret Agent container. Confirmed this fails against the pre-fix code with `x509: decryption password incorrect` and passes with the fix.
- `pkg/service/secret/cluster_tls_test.go` — resolver unit tests: resolution through the cluster's agent, all other TLS fields preserved, cluster left unmutated, no agent call when there is nothing to resolve, error propagation.
- `pkg/service/aerospike/client_factory_test.go` — new file; covers resolution wiring, resolution-failure propagation, `NewTLSConfig`-failure propagation, and the no-TLS-required early exit.

**Not included:** `.gitignore` picks up a one-line unrelated addition committed on this branch.

[BKRS-394]: https://aerospike.atlassian.net/browse/BKRS-394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


filkeith (approved):
lgtm

PR #662, opened 2026-09-13
BKRS-XXX Hand out the run context in Components.Run, not during InitComponents

korotkov-aerospike:
## What does this change do?

`InitComponents` wired the object graph but also let the build context escape from it, in two places:

- `ApplyNewConfig(ctx)` launched the storage history scan on it — a goroutine reading S3/GCS/Azure before the caller had decided to start anything.
- `handlers.Service` stored it as `sysCtx //nolint:containedctx` and replayed it to start restores and later re-applies.

Both needed a *service lifetime* and had no moment to receive one, so they took whatever context was in reach. `Components.Run` is now that moment; `main` is build-then-`Run`.

- **History sync is a queue.** `BackupStateRegistry.RequestHistorySync(routines)` coalesces per routine and returns at once; `Start(ctx)` serves the queue. The scan `ApplyNewConfig` requests during `InitComponents` therefore cannot run before the service does — there is no context to run it on. `ApplyNewConfig()` loses its ctx parameter.
- **Restore lifetime lives in `RestoreJobsHolder`.** `Start(ctx)` sets the parent of every job; `Restore(req)` / `RestoreByTime(req)` drop the ctx that only ever named that parent. Before `Start`, `newJob` returns `ErrRestoreNotStarted` (unreachable in production — the listener isn't up yet).
- **`handlers.Service` stores no context.** The one remaining `containedctx` suppression is in the holder, set at `Start` rather than construction, with a paragraph saying why.

The rule this leaves behind, stated in the `InitComponents` doc comment: the build context is for loading (config, TLS material, secrets); nothing may capture it or launch on it.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Chore / refactor / CI

## Checklist

- [ ] Target branch is `dev`. Only release PRs and hotfixes target `main`; v2 maintenance targets `v2`. — **Stacked on `rate-limiter-lazy-eviction`** (needs the ctx-free server constructors); retarget as the stack merges.
- [x] `make pr` passes locally (tidy, mocks, format, lint, tests, docs). — tidy, mocks regenerated, format, lint, `-race` tests; no REST API change so no doc regeneration.
      [migration guide](../README.md#migration-guide) is updated. — not applicable; internal Go API only.

## Additional context

Behavior changes to be aware of:

- Scans are serialized: each `ApplyNewConfig` used to spawn its own, so two quick applies could scan the same routine concurrently. Now one scan runs at a time and requests arriving mid-scan run after it.
- `finishJob` now releases every job's context. The time runner previously never canceled its own (a small leak); the path runner's ad-hoc `cancel()`-on-error is subsumed.
- Restore jobs end with the run context. A `Stop()` that also waited for them was considered and dropped: `Scheduler.Stop()` doesn't wait for in-flight backups either, and draining one but not the other would be inconsistency dressed as care.

New tests: `TestComponents_RunServesUntilCanceled` drives the real lifecycle through the entry point `main` uses; the registry queue and holder lifetime each have focused unit tests. The mock choreography around `SynchroniseBackupHistory` collapsed into synchronous expectations.

Not in this PR, by design: fail-on-call fakes at the composition root (needs an injection seam for storage/client factories), a `New*`-has-no-`go` analyzer, and import-boundary tests.


korotkov-aerospike:
Superseded by three smaller PRs, each with its own ticket — the change is identical (verified byte-for-byte against this branch's tree, rebased onto the current #661 tip):

- #665 — [BKRS-400](https://aerospike.atlassian.net/browse/BKRS-400) Introduce `Components.Run` as the service lifecycle entry point (pure move, no behaviour change)
- #666 — [BKRS-401](https://aerospike.atlassian.net/browse/BKRS-401) Queue backup-history sync in the registry instead of launching it from config apply
- #667 — [BKRS-402](https://aerospike.atlassian.net/browse/BKRS-402) Give `RestoreJobsHolder` the restore job lifetime, drop the stored context from `handlers.Service`

Each builds, lints and passes `-race` tests on its own.

[BKRS-400]: https://aerospike.atlassian.net/browse/BKRS-400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BKRS-401]: https://aerospike.atlassian.net/browse/BKRS-401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BKRS-402]: https://aerospike.atlassian.net/browse/BKRS-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #664, opened 2026-09-14
BKRS-398 Type model credential fields as Secret to stop log leaks

korotkov-aerospike:
## What does this change do?

Fixes a credential leak in JSON-format logs ([BKRS-398](https://aerospike.atlassian.net/browse/BKRS-398)). Redaction was keyed on a `Secret` type that only the DTO layer used, and the DTO to model conversion turned every credential into a plain `string`. Business logic logs model structs, for example `slog.Any("backup", details)` in the restore runners, and the JSON handler marshaled `Storage.Auth` verbatim. Plain format only survived because the storage types' `String()` omits auth, which was incidental rather than by design.

**The type.** `Secret` lives in its own leaf package, `pkg/redact`, since `pkg/model`, `pkg/dto` and the secret resolver all need it. It is named `redact` rather than `secret` because `pkg/service/secret` already exists and `secret.Secret` would stutter.

**The fields.** Every model credential field is now `redact.Secret`: S3 key ID and access key, Azure account key, tenant ID, client ID and client secret, GCP key JSON, cluster password, encryption key secret, cluster TLS key-file password and HTTPS server key-file password. The existing `ReplaceAttr` walk and `String()` then redact model structs in both log formats with no per-call-site work, and the DTO to model conversions lose their `string()` and `secret()` wrappers.

**Reading a secret.** `Secret.Reveal()` is the single sanctioned way to get the literal, so grepping for it finds every point a credential leaves the type. `secrets.Resolver.Resolve` takes a `redact.Secret`, making the resolver the main reveal point; the others are the two `scan.go` files handing the encryption key to backup-go, and the TLS key decryption.

**Hashing.** `Secret.Hash()` is used by `Credentials.Hash` and `TLS.Hash`. `hashValues` formats with `%v`, so without it every literal secret would collide on the `[secret]` placeholder and a password rotation would silently keep serving a cached client.

**The walks key on an interface.** `RedactSecrets` and `MergeSecrets` matched `reflect.TypeFor[redact.Secret]` exactly, so any second credential-bearing type would be walked straight past, with no redaction in API responses and no `[secret]` preservation on merge. They now test for a `redact.Redactable` interface, and build the replacement as `reflect.New(v.Type())` plus `SetString` so it keeps the value's own type instead of assuming `redact.Secret`.

Also removes the unused `AerospikeCluster.GetPassword` accessor, and adds `pkg/redact` to the swagger generator's search path so it can resolve the field type. Generated docs are unchanged.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Chore / refactor / CI

## Checklist

## Additional context

No API or config change: the DTO field types were already the same string-kind type under a different name, and generated docs are unchanged.

Tests worth looking at:
- `TestHandlerReplaceAttr_RedactsModelStorageCredentials` runs S3, Azure shared key, Azure AD and GCP storages through both JSON and plain handlers, for a single `BackupDetails` and a slice, asserting the literal never appears.
- `TestRedactSecrets_AnyRedactableType` and `TestMergeSecrets_AnyRedactableType` use a second, unrelated secret-bearing type to prove both walks key on the interface and preserve the value's own type.
- `TestSecret_Hash` and `TestTLS_Hash` cover the hashing.

Follow-ups noticed during review, not addressed here:
- `AzureStorage.Endpoint` stays a plain string although SAS-bearing endpoint URLs are supported, so a SAS signature in the endpoint would still reach logs.
- `Resolver.Resolve` returns a plain `string`, so the resolved literal is stored untyped on backup-go config structs.
- `AerospikeCluster.Hash` feeds pointer fields to a `%v`-based hasher, so it hashes addresses rather than values. Pre-existing.
- `build/scripts/generate-mocks.sh` does not cover `pkg/service/secret`, because it derives the package name from the directory and that package is named `secrets`. The mock here was regenerated with the command in its own header.

[BKRS-398]: https://aerospike.atlassian.net/browse/BKRS-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


filkeith (approved):
lgtm

PR #667, opened 2026-09-14
BKRS-402 Give RestoreJobsHolder the restore job lifetime, drop the stored context from handlers.Service

korotkov-aerospike:
Jira: [BKRS-402](https://aerospike.atlassian.net/browse/BKRS-402)

## What does this change do?

`handlers.Service` stored the application context as `sysCtx context.Context //nolint:containedctx`, captured at construction, and replayed it to start restore jobs. Two problems: an HTTP adapter should not decide how long a database restore lives, and a context captured while the graph is being built is live from the first line of `InitComponents` — so a context capable of launching work existed long before the service was running.

`RestoreJobsHolder` already held every job's `cancel` func, so the service lifetime is just the parent of the contexts it already manages. It receives that lifetime in `Start(ctx)`, and `newJob` returns the context the job must run in.

`Restore(request)` and `RestoreByTime(request)` therefore lose the `ctx` parameter that never meant what its signature implied: callers could not influence a job's lifetime, and the handlers deliberately passed something other than `r.Context()`. They also lose their `error` return, which both implementations had always hard-coded to `nil` — a restore request is validated in the DTO layer, and everything that can go wrong after that is reported through `JobStatus`. Both handlers collapse to one line, and every handler now has exactly one context in scope — the request's — so there is no second one to pick wrongly.

With BKRS-401 already queuing the history sync, nothing in the HTTP layer holds a context any more, and the rule is stated in the `InitComponents` doc comment: **the build context is for loading; nothing may capture it or launch work on it.** The one remaining `containedctx` suppression in the repository is in the holder, set at `Start` rather than at construction, with its reason written down.

Two things make that rule checkable rather than merely asserted:

- `TestInitComponents_BuildContextIsNotALifetime` builds the graph, **cancels the build context immediately**, then serves on an independent one and requires the service to come up. Pointing `Run` at the cancelled context makes it fail, so it has teeth.
- Auditing the graph against the rule turned up one violation. quartz's `SlogLogger` stores the context it is handed for the life of the scheduler and passes it to every log call — which is the only reason `NewScheduler` ever took a `ctx`. quartz's logger interface has no contexts in it, so `schedulerLogger` adapts `*slog.Logger` directly and the parameter goes away. `backup_scheduler.go` no longer imports `context` at all, and the build context is now used at exactly two call sites, both genuine loading.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Chore / refactor / CI

## Checklist

- [ ] Target branch is `dev`. Only release PRs and hotfixes target `main`; v2 maintenance targets `v2`. — **stacked on BKRS-401**; retarget as the stack merges.
      [migration guide](../README.md#migration-guide) is updated. — not applicable; internal Go API only, no REST change.

## Additional context

Bug fixed along the way (hence "Bug fix" ticked): a completed restore never released its job context. The restore-by-time runner never cancelled at all, and the path runner cancelled only on failure — so every successful path restore, and every restore-by-time either way, left its context registered in the service context's `children` map until the process exited. `finishJob` now releases the context however the job ended, which also subsumes the path runner's ad-hoc `cancel()`-on-error. `go vet`'s `lostcancel` cannot catch this: `cancel` *was* reachable on one path, which is all that check asks for.

[BKRS-402]: https://aerospike.atlassian.net/browse/BKRS-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -81,6 +81,12 @@ linters:
           msg: use decoder.Marshal instead
         - pattern: '^yaml\.NewDecoder$'
           msg: use decoder.Deserialize instead
+        - pattern: '^time\.Sleep$'
+          msg: 'not cancellable; use select { case <-time.After(d): case <-ctx.Done(): }'
+        - pattern: '^context\.Background$'
+          msg: pass the caller's context, or context.WithoutCancel(ctx) for cleanup after cancellation
+        - pattern: '^context\.TODO$'
+          msg: pass the caller's context, or context.WithoutCancel(ctx) for cleanup after cancellation
     nlreturn:
       block-size: 10
     misspell:
@@ -122,6 +128,11 @@ linters:
       - linters:
           - forbidigo
         path: pkg/dto/decoder/
+      # main owns the root context: signal.NotifyContext(context.Background(), …)
+      # is where cancellation starts, so it is the one place that may create one.
+      - linters:
+          - forbidigo
+        path: cmd/backup/main\.go
       - linters:
           - forbidigo
         path: build/
@@ -140,6 +151,8 @@ linters:
           - unparam
         path: _test\.go
     paths:
+      # zz_*_test.go are throwaway audit scratch files, not part of the suite.
+      - zz_.*_test\.go$
       - examples
       - modules
       - docs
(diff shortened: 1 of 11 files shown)


filkeith (approved):
lgtm

PR #669, opened 2026-09-14
BKRS-421 Fix the trivial pre-release audit findings and keep the audit tests

korotkov-aerospike:
## What does this change do?

Fixes every finding of the independent pre-release code audit that had a trivial fix, and keeps the audit's tests as regression tests, named after what they check: each fixed finding's test asserts the corrected behaviour; tests for the findings that are **not** fixed here assert the expected behaviour and are skipped with `t.Skip("BKRS-nnn: ...")` until that ticket is fixed, so nothing in the tree asserts a defect and nothing is red.

Umbrella ticket: [BKRS-421](https://aerospike.atlassian.net/browse/BKRS-421). Non-trivial findings were opened as BKRS-405 … BKRS-420; documentation findings are tracked in BKRS-404 and BKRS-397.

**Fixed here**

- `errNotFound` printed numeric restore job IDs as rune literals (`job 'A' not found`); `statusCodeError` now unwraps its cause so `errors.Is/As` see through it.
- Ad-hoc backup `delay` values that overflow a `time.Duration` wrapped negative and ran the backup immediately; now 400.
- `PUT /v1/config` merges secrets **before** the static-field comparison, so a GET-edit-PUT round trip carrying a redacted `service.https.key-file-password` is accepted instead of failing with "KeyFilePassword changed".
- Every JSON endpoint now goes through `decodeBody`: an empty or whitespace-only body is 400 (an empty `PUT /v1/config` used to wipe the whole configuration with 200) and bodies above 8 MiB are 413.
- JSON `Deserialize` rejects data after the first document.
- `"backup-routines": {"x": null}` panicked in `Config.Validate`; a null routine or a null secret-agent entry is a validation error (the null agent used to be resolved, by nil-pointer equality, as the agent of every agent-less entity).
- Incremental-cron validation error quoted the full cron instead of the incremental one.
- Partial Azure auth (`account-name` without `account-key`, AAD without `client-secret`) validated and silently became anonymous access; now rejected.
- `PUT /v1/config/clusters/{name}` and `/policies/{name}` invalidate the routines using them, so scheduled jobs pick up the change (they hold a routine snapshot).
- A cached Aerospike client that fails its health check is closed and dropped, so the next call reconnects instead of reusing the dead client forever.
- Retry back-off saturates at `math.MaxInt64` instead of overflowing (negative on amd64 = hot loop).
- On shutdown the partial-backup cleanup (`OnFail`) received an already-cancelled context and never deleted the folder; it now gets `context.WithoutCancel`.
- Root endpoint returned 404 under a non-default `context-path`; it now derives the root from the matched pattern.
- Restore-by-timestamp checked emptiness of the *source* namespace even when the policy remaps to a destination; it now checks the destination.
- `log.NewHandler` probes the log file at startup; an unwritable file fails the bootstrap instead of leaving the service running with no logs.
- Restore retry-policy default described as "one-minute delay" in the DTO comment (rendered into docs); corrected.

**Test housekeeping**

- Deleted: tests already covered elsewhere (listener port conflict, storage-relative path traversal, CRL freshness sentinels, HTTPS defaults), documentation-default tests now enforced by `internal/doccheck`, and documentation-mismatch tests tracked in BKRS-404 / BKRS-397.
- Kept as regression tests: model credential redaction in both log formats (BKRS-398), CRL trust-anchor rejection, rate limiter ignoring forwarded headers, storage error formatting, restore-by-time boundary and chain selection, idempotent retention delete.
- Two existing tests that enshrined fixed bugs were updated: `api_contract_test.go` (`job '\x01' not found`) and `Test_GetClient_UnhealthyConnection` (dead client must be closed).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Chore / refactor / CI

## Checklist

      [migration guide](../README.md#migration-guide) is updated. — not applicable: no format change; the API only becomes stricter (see below).

## Additional context

Behaviour visible to API clients and operators: empty bodies are 400 and bodies over 8 MiB are 413; a partial Azure auth block, a null routine and a null secret-agent entry are rejected at validation; trailing data after a JSON body is rejected; an unwritable `file-writer.filename` fails startup; numeric job IDs in "not found" messages are printed as numbers.

GoLand still reports IDE-level "duplicated code fragment" weak warnings for the two restore handlers and the routine enable/disable handlers; they are structurally parallel handlers and were flagged before this change as well.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[BKRS-421]: https://aerospike.atlassian.net/browse/BKRS-421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


filkeith (approved):
lgtm

PR #676, opened 2026-09-15
BKRS-414 Validate configuration names as single path segments

korotkov-aerospike:
## What does this change do?

Fixes [BKRS-414](https://aerospike.atlassian.net/browse/BKRS-414) (audit severity: **high**).

`dto.Path` was never applied to entity names or namespaces — `Config.Validate` checked them only for non-emptiness — while `PathService` joins `routineName` and `namespace` into storage paths verbatim. A routine named `../../escaped-routine` with namespace `../../escaped-ns` on local storage `/var/backups/abs` wrote to `/var/escaped-routine/backup/escaped-ns`, and retention with `incremental: 0` ran a recursive `RemoveFiles` on `/var/escaped-routine/incremental`. On S3/GCP/Azure it escaped the configured key prefix. The name arrives through `POST /v1/config/routines/{name}` — the mux decodes `..%2F..%2Fescape` into `PathValue` — or through any YAML key.

Validation now happens in the DTO layer, where the rest of the configuration is validated.

### `NamespaceName`, modelled on `Path`

A namespace is not only a name in the cluster: it is also the folder a routine's data for that namespace is written to. It gets a type that validates itself, the way `Path` does:

```go
type NamespaceName string

func (n NamespaceName) Validate() error
```

`Validate` reports only the reason — `errEmpty` for a missing name, a plain error otherwise — and the call site wraps with `errValidationInvalidName`, which supplies the field name and the offending value and maps `errEmpty` to `errValidationEmptyField`. That mirrors `Path.Validate` / `errValidationInvalidPath` exactly, so both name checkers and the path checker classify errors the same way.

The rules are the [Aerospike naming rules](https://aerospike.com/docs/database/reference/limitations/#namespace): at most 31 bytes of Latin letters, digits, `_`, `-` and `$`, and not the reserved name `null`. They are stricter than a single path segment has to be, so they cover the storage layout as well.

It is the **field type** wherever a namespace is named:

- `BackupRoutine.Namespaces` is `*[]NamespaceName`;
- `RestoreNamespace.Source` and `.Destination` are `NamespaceName`.

The model keeps plain `[]string`; `toModel` / `fromModel` convert. A side effect worth noting: `fromModel` used to do `r.Namespaces = &m.Namespaces`, handing the DTO a pointer to the model's own slice — it now copies.

### Entity names

The key of a routine, storage, cluster, policy or secret agent must be a single path segment: not `.` or `..`, no `/` or `\`, no NUL byte, no leading `~`. These stay a function (`validateEntityName`) rather than a type, because they are **map keys** — giving them a type would mean changing the key type of every configuration map and its JSON/YAML marshalling for no validation benefit.
- **Entity names get the minimum rule that closes the hole, not a character whitelist.** These names are user-chosen labels; `cluster.prod`, `my routine` and `backup_2024` are all reasonable and all still accepted. A whitelist would reject working configurations on upgrade for no security gain — a name containing no separator cannot leave its folder.
- **Namespaces get the full Aerospike rule.** A namespace name is not free-form: it names something that has to exist in the server, and the server's own documented limits are stricter than a path segment needs. A name outside that set is one the user could not have been backing up successfully. The caveat is that Aerospike says the guideline "is not enforced", so a name that works on the server but is not documented (a `.`, say) would now be rejected. That is the deliberate trade; the alternative — accepting anything without a separator — leaves 31-byte and `null` mistakes to surface much later, as a failing backup.
- **`RestoreNamespace` is validated too.**  One rule for one concept is the point of the type, but it can reject a restore of a backup whose namespace name the server tolerated undocumented.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Chore / refactor / CI

## Checklist

- [x] If this changes the configuration file or REST API in a breaking way, the [migration guide](../README.md#migration-guide) is updated.

## Additional context

The named type does not leak into the generated artifacts — `namespaces` is still `{"type": "array", "items": {"type": "string"}}` and `source` still `{"type": "string"}`, exactly as `Path` behaves. The only generated-doc change is the naming-rules prose on the three namespace fields.

[BKRS-414]: https://aerospike.atlassian.net/browse/BKRS-414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/dto/validate_name.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if len(n) > maxNamespaceNameLength {
+		return fmt.Errorf("must not exceed %d bytes", maxNamespaceNameLength)
+	}
+
+	if n == reservedNamespaceName {
+		return fmt.Errorf("%q is reserved by the Aerospike server", reservedNamespaceName)
+	}
+
+	for _, r := range n {
```
filkeith: I think you should replace it with regexp? Going through each letter in a string looks strange

filkeith (changes requested):

filkeith (approved):

PR #677, opened 2026-09-15
BKRS-407 Fail a routine run when a namespace backup cannot start

korotkov-aerospike:
## What does this change do?

Fixes [BKRS-407](https://aerospike.atlassian.net/browse/BKRS-407) (audit severity: **high**).

`waitUntilBackupStarted` polled `GetStats() == nil` and nothing else. When `Start()` failed permanently — cluster unreachable, storage writer that cannot be created — the retry loop finished and recorded the error, but the statistics stayed nil, so `Run` blocked until the scheduler's long-lived context was cancelled at shutdown. The consequences compounded:

- the `StartController` reservation was held the whole time, so every later scheduled or ad-hoc backup of that routine was skipped with "backup already in progress";
- `runBackupInternal` never returned, so `outcomeReporter.Report` never ran — no failure metric, no Error log;
- a source cluster that was down at one scheduled time stopped all future backups of that routine until a restart.

The fix asks the thing that knows the answer instead of polling a side effect:

- `retryableBackupHandler` gains `started`, closed once the pipeline is running, and `done`/`err`, the retry loop's outcome. `waitStarted` selects over the two and the caller's context.
- The constructor becomes `startRetryableBackup`, which returns `(*retryableBackupHandler, error)`: a handler once the pipeline is live, or the error that ended the run and nothing left running. A run that never starts one has nothing to observe, so there is no handler to hand back.
- `NamespaceBackupRunner.Run` returns `(CancelableBackupHandler, error)` and does that waiting itself. `routineBackupRunner` starts the namespaces through an `errgroup` and reports the first failure; the orchestrator then reports it the way it reports any other failure, and releases the reservation.
- `Wait` reads `done`/`err` rather than consuming a single-slot channel, so it is now idempotent — `BackupNamespacesOperation.Wait` calls it per namespace.

A run that gives up also cancels the namespaces that *did* start. They shared the abandoned run's timestamp folder and had nobody left to wait for them, so leaving them running would keep writing data into a folder that will never be completed or reported.

### Judgement calls

- **The "started" condition is now "the pipeline handler is set" rather than "the first non-nil `GetStats()`".** The two are equivalent in practice — `backupexecutor.BackupHandler` documents that backup-go handlers never return nil statistics once constructed — but the new one is a signal rather than a 100 ms poll of a side effect, and it can distinguish "not yet" from "never".
- **Retries still span the start.** `waitStarted` returns an error only once the whole retry loop has given up, so a transient start failure is retried before the run is failed. One attempt is `Start` → `Wait` → metadata write, and the retry budget covers all of it.
- **A run that finishes between starting and the wait is a started run**, not an error: its result belongs to `Wait`, so a routine run is reported the same way whenever it fails.
- **A namespace that cannot start fails the whole routine run**, rather than proceeding with the ones that did. The namespaces of one run share a timestamp folder and are restored as a set, so a partial run is not a useful backup.
- **Cancelling the started namespaces is an addition to the ticket**, but it follows directly: without it, fixing the hang would have turned a blocked run into a leak.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Chore / refactor / CI

## Checklist

- [x] If this changes the configuration file or REST API in a breaking way, the [migration guide](../README.md#migration-guide) is updated.

## Additional context

`MockNamespaceBackupRunner` is regenerated for the new `Run` signature; no type was added to the mock list. `BackupNamespacesOperation` is unchanged — the aggregate has no single run to report on.

A start failure deliberately does not call `completionHandler.OnFailure`. That only clears the registry entry made by `BackupStarted`, which never ran for a run that never started.

Tests:

- `pkg/service/retryable_backup_handler_test.go` — a run that never starts fails the call itself and yields no handler; a run that started and then failed is accepted, with the failure left to `Wait`; the wait is bounded by the caller's context; a table test pins `waitStarted` across the four `started`/`done` states, including both orders of the race between them.
- `pkg/service/routine_backup_runner_test.go` — a namespace that cannot start fails the run well inside a 700 ms bound, both with a mocked runner and end-to-end through the real `retryableBackupHandler` with an executor that fails to start; the namespaces that started are cancelled when another one fails.

`go test -race ./pkg/service/...` passes, including a `-count=30` pass over the handler tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[BKRS-407]: https://aerospike.atlassian.net/browse/BKRS-407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/service/retryable_backup_handler.go
```diff
@@ -81,27 +95,85 @@ func newRetryableBackupHandler(
...
+		}
+
+		return h.result()
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// finish records the run's outcome and releases everyone waiting on it.
+func (h *retryableBackupHandler) finish(err error) {
```
filkeith: May be it make sense to make errorChan and send error there. So you can get rid of mutexes, and check error in the same select in waitStarted()

filkeith (changes requested):

pkg/service/retryable_backup_handler.go
korotkov-aerospike: You're right about the mutex — dropping it. But it's the `close` that makes it safe rather than moving the payload into the channel: `h.err` is written before `close(h.done)`, and both readers only touch it inside a `case <-h.done:` branch, so the close publishes the write. The lock was belt-and-braces.

The error can't move into the channel though, because the outcome now has two readers: `waitStarted` asks *did it ever start*, and `Wait` asks *how did it end* — the latter once per namespace via `BackupNamespacesOperation.Wait`. A value sent on a channel goes to exactly one receiver; only `close` broadcasts. With `errCh` the first reader would take the error and the second would block until its context expired, which for a scheduled backup is shutdown — the same hang this PR is fixing, moved one level down. That's why `errCh` worked before: the old `waitUntilBackupStarted` polled `GetStats()` and never read it, so there was exactly one reader.

So it ends up as the `context.Context` shape — a channel to broadcast that it happened, a field for what happened.

On checking it in the same select: that part already works — `<-h.done` is the second case in `waitStarted`, which is how a run that never starts a pipeline reports its error instead of waiting for statistics that will never arrive.

pkg/service/retryable_backup_handler.go
```diff
@@ -81,27 +96,69 @@ func newRetryableBackupHandler(
...
+// signal waited for here is produced by the loop started here. The run's wait context is released
+// as soon as that loop returns, so a finished run leaves nothing behind in the scheduler context.
+func (h *retryableBackupHandler) start(ctx context.Context, run func() error) error {
 	go func() {
-		defer cancel()
-		h.errCh <- try.Retry(ctxWithCancel, policy,
-			logger.With(slog.String("label", "backup")), processBackup, callbacks.OnRetry)
+		defer h.cancel()
+
+		h.err = run()
```
filkeith, line 120: error is not protected by mutex or smth like this. Is it ok? you then read it in 2 places

filkeith (changes requested):

filkeith (approved):

PR #686, opened 2026-09-22
BKRS-439 config mutation status codes

korotkov-aerospike:
## What does this change do?

Fixes [BKRS-439](https://aerospike.atlassian.net/browse/BKRS-439). Every config-mutation endpoint answered `400 Bad Request` for a resource that does not exist, while `GET` on that same resource correctly answered `404`. A client could not tell "you sent nonsense" from "that entity is not there" by status code, and had to string-match the response body instead.

```
GET    /v1/config/routines/doesNotExist  -> 404 routine "doesNotExist" not found
DELETE /v1/config/routines/doesNotExist  -> 400 invalid request: failed to update configuration:
                                                delete backup routine "doesNotExist": item not found
```

`httpError` already mapped `model.ErrNotFound` to 404, but these handlers wrapped *every* error out of `changeBackupConfig` in `errBadRequest(...)`, which produces a `*statusCodeError{Code: 400}`. `httpError`'s switch tests `errors.As(&statusCodeError)` before `errors.Is(err, model.ErrNotFound)`, so the 404 branch was unreachable from those call sites even though the error underneath did wrap the sentinel.

### Status codes

| Condition | Before | After |
| --- | --- | --- |
| `DELETE` / `PUT` a routine, storage, cluster or policy that does not exist | 400 | **404** |
| `POST` a name that is already taken | 400 | **409** |
| `DELETE` a storage, cluster or policy a routine still references | 400 | **409** |
| Malformed body, failed validation, unreachable TLS endpoint | 400 | 400 (unchanged) |

404 bodies now read exactly as the matching `GET` does — `routine "x" not found` — rather than the nested `invalid request: failed to update configuration: ...` string. The 409 in-use body keeps the `it is used in routine "y"` detail.

### How

- **`errConfigChange(err, entity, name)`** (`internal/server/handlers/errors.go`) maps the three `model` sentinels onto statuses, and is now the single error path for all 14 config-mutation handlers, replacing the blanket `errBadRequest`. Anything that is not one of the three sentinels still falls through to 400, so validation and TLS-probe failures are unaffected.
- **`dto.Config` gained `Add`/`Update`/`Delete` per entity type**, plus `SetRoutineDisabled`. Each returns `([]string, error)` — the routines the change invalidates — which is exactly the `changeBackupConfig` mutate signature, so every handler body collapsed to one line:

  ```go
  err := s.changeBackupConfig(r.Context(), func(config *dto.Config) ([]string, error) {
      return config.DeleteRoutine(name)
  })
  ```

  This mirrors the `AddStorage`/`DeleteStorage`/`DeleteCluster` API `model.Config` already exposes.
- **The in-use checks moved into the DTO.** `ensurePolicyNotInUse`/`ensureClusterNotInUse`/`ensureStorageNotInUse` and the three `routinesUsing*` helpers are gone from `handlers/config_change.go`; they are now unexported details of the `Delete`/`Update` methods that need them.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Breaking change (config file, REST API, or CLI flags)
- [ ] Documentation
- [ ] Chore / refactor / CI

## Checklist

- [x] `make pr` passes locally (tidy, mocks, format, lint, tests, OpenAPI, README generation).
- [x] Generated artifacts are committed and up to date (`make generated-check` passes): mocks, OpenAPI spec, README/docs.

## Additional context

**`ErrInUse` → 409 is a deliberate widening.** BKRS-439 names only `ErrNotFound` (→ 404) and raises `ErrAlreadyExists` as "worth confirming during the fix". `ErrInUse` is not mentioned in the ticket, but after this change it is the only conflict-shaped outcome left on 400, and a `DELETE` of a storage that a routine still references is a well-formed request conflicting with current state — the textbook 409. The resulting rule is uniform: **404 missing, 409 conflict, 400 malformed.**

**One deliberate override of a guardrail.** `TestUnconfiguredService_APIContract` carries the comment *"A failure here is a release blocker: fix the regression rather than updating the expectations below."* Five rows were updated (DELETE cluster/storage/policy/routine, PUT policy) because those expectations *are* the bug this PR fixes. Worth a reviewer's eye to confirm that reading.

**Unrelated work is bundled on this branch.** Alongside the BKRS-439 commits it also carries a `backup-go` bump (`20260907091536` → `20260922102415`), a new `estimates` doc-tag generator in `build/docs/readme_generator.go`, the corresponding `docs/monitoring.md` tag, and doc-comment edits in `backup_policy.go` / `restore_policy.go`. These are independent of the status-code change and may be better split out.

**Tests.** `internal/server/handlers/config_change_status_test.go` pins the add-conflict / update-missing matrix across all four entity types (8 cases). `pkg/dto/config_change_test.go` covers the new DTO methods directly, including the invalidation names the handler tests cannot observe. The existing handler tables and `TestUnconfiguredService_APIContract` were updated to the new codes.

**Docs.** `CHANGELOG.md` under `3.7.0 / Changed`, and `docs/migration.md` under `v3.6 -> v3.7 / Breaking changes` with the before/after table and the client-side impact.

[BKRS-439]: https://aerospike.atlassian.net/browse/BKRS-439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


filkeith (dismissed):
LGTM, update doc about validation name changes

filkeith (approved):

PR #688, opened 2026-09-23
BKRS-438 Make filter-exp validation honest and stop retrying permanent failures

korotkov-aerospike:
## What does this change do?

[BKRS-438](https://aerospike.atlassian.net/browse/BKRS-438) reports that a `filter-exp` holding a random base64 string is accepted, while the rejection message (`failed to parse filter expression`) implies a parse that never happens.

**1. The config-time check now says what it does.** It still runs through the Aerospike client's own constructor, so which encoding `filter-exp` uses stays the client's business rather than this package's; the constructed expression is discarded, since the call is made for the check alone. The message now says *decode* rather than *parse*, because decoding is all that happens. The payload itself is still not parsed — see Additional context for why that is not available to us — so this does not stop a garbage expression being accepted. A skipped test referencing the ticket pins that gap instead of leaving it implicit.

**2. Such an expression is no longer retried.** Because it does reach the cluster, the cluster rejects it, and `try.Retry` decided retries from context alone: every scheduled run spent its whole attempt budget, back-off included, to reach the identical failure. Failures are now classified:

- **By backup-go error class.** The library publishes these as part of its contract and documents `ErrInvalidConfig`, `ErrCorruptData` and `ErrNotFound` as pointless to retry. ABS did not use `errclass` at all before this. The one entry this replaces, `asinfo.ErrNoNode`, is itself `fmt.Errorf("%w: no node found", errclass.ErrNotFound)`, so it is subsumed rather than dropped.
- **By Aerospike result code.** The class alone is not enough: `ErrAerospike` covers both a timeout worth retrying and a malformed request that never will be. backup-go wraps the client error rather than replacing it (`record_reader.go:250` uses two `%w` verbs), so the code stays reachable via `errors.As`.

A filter expression the server cannot parse comes back as `PARAMETER_ERROR`, **measured against a real cluster**, not assumed. The new integration test asserts both halves — that the cluster answers `PARAMETER_ERROR`, and that `try.Retry` then attempts exactly once — so a future server version cannot change this unnoticed.

This PR also carries three unrelated refactoring commits that were ready and too small to ship alone:

| Commit | Change |
| --- | --- |
| `cluster label` | `AerospikeCluster.Label()` replaces the label-or-seed-node-or-unknown fallback repeated across 5 call sites |
| `refactor RestoreValidator` | Separates fact-gathering from rule application, adds `RestoreNamespace.DestinationOr` and `BackupRoutine.BacksUpWholeCluster`, validates that a namespace remap is scoped to exactly its source namespace, and removes two unreachable guards |
| `Move running/empty state checks into model types` | `RoutineState.IsRunning()` and `BackupDetails.IsEmpty()` replace pointer and field inspection at the call sites |

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Chore / refactor / CI

## Checklist

- [ ] If this changes the configuration file or REST API in a breaking way, the [migration guide](../README.md#migration-guide) is updated.

[BKRS-438]: https://aerospike.atlassian.net/browse/BKRS-438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/util/collections/missing.go
```diff
@@ -15,3 +15,32 @@ func MissingElements(subset, superset []string) []string {
 
 	return missing
 }
+
+// FirstCommon returns the first element of b that also appears in a, or ("", false) if none.
+func FirstCommon(a, b []string) (string, bool) {
+	for _, v := range b {
+		if slices.Contains(a, v) {
```
filkeith, line 22: Not optimal algorythm O(len(a) × len(b))
If you have a small amount of elements ok, but my be it make sense to rewrite it in the future

pkg/util/try/retry.go
```diff
@@ -9,21 +9,63 @@ import (
...
 	"github.com/aerospike/backup-go/models"
-	"github.com/aerospike/backup-go/pkg/asinfo"
 )
 
-var nonRetryableErrors = []error{asinfo.ErrNoNode}
+// nonRetryableClasses are the backup-go error classes whose cause cannot change between
+// attempts. The library documents the classification as part of its contract, so matching on
+// these is stable in a way that matching on individual errors or messages is not.
+var nonRetryableClasses = []error{
+	errclass.ErrInvalidConfig,
```
filkeith, line 22: there are links in backup-go root https://github.com/aerospike/backup-go/blob/dev/errors.go so you do not need to import errclass just backup.ErrInvalidConfig

filkeith (dismissed):
LGTM, take a look at 2 minior comments

pkg/util/collections/missing.go
korotkov-aerospike: We don't care about the speed here, but I renamed it.

filkeith (approved):

